PR #7, opened 2020-11-24
Allocator visualizer

max-traverse:
Introducing an optional feature: the allocator visualizer.
It's an imgui based UI that allows you to inspect where each allocation is ending up. It can even show the backtrace of each allocation that happened.
It also allows the user to inspect information about the different types of memory available on the GPU.


examples/vulkan-visualization/src/main.rs
```diff
@@ -0,0 +1,1417 @@
...
+            let app_name = CString::new("gpu-allocator examples vulkan-visualization").unwrap();
+
+            let appinfo = vk::ApplicationInfo::builder()
+                .application_name(&app_name)
+                .application_version(0)
+                .engine_name(&app_name)
+                .engine_version(0)
+                .api_version(vk::make_version(1, 0, 0));
+
+            let layer_names = [CString::new("VK_LAYER_KHRONOS_validation").unwrap()];
```
Jasper-Bekkers, line 1048: Validation layer should probably be optional / off by default since not everybody will have the VulkanSDK installed. (Or at least print out a message where to download the SDK if it fails).

examples/vulkan-visualization/src/main.rs
```diff
@@ -0,0 +1,1417 @@
...
+
+        let mut imgui_renderer = ImGuiRenderer::new(
+            &mut imgui,
+            &device,
+            descriptor_pool,
+            &mut allocator,
+            setup_command_buffer,
+            setup_commands_reuse_fence,
+            present_queue,
+        )?;
```
Jasper-Bekkers, line 1286: Since we have so little code that actually demos the example it might be nicer to pull it out to a loose file and leave all the ash/vulkan extra code in a helper file (might be useful for some other examples later on too).

examples/vulkan-visualization/src/shaders/imgui.ps.hlsl
```diff
@@ -0,0 +1,15 @@
...
+struct VertexInput
+{
+    float4 position : SV_POSITION;
+    float2 texCoord: TEXCOORD0;
+    float4 color: COLOR;
+};
+
+float4 main(VertexInput input) : SV_Target0
+{
+    //return input.color;
```
Jasper-Bekkers, line 13: Nit: commented code.

examples/vulkan-visualization/src/shaders/imgui.vs.hlsl
```diff
@@ -0,0 +1,32 @@
...
+};
+
+ConstantBuffer<Constants> g_constants : register(b0, space0);
+
+VertexOutput main(VertexInput vertex)
+{
+    VertexOutput o;
+    o.position = float4(vertex.pos * g_constants.scale + g_constants.translation, 0.0, 1.0);
+    //o.position = float4((vertex.pos + g_constants.translation) * g_constants.scale, 0.0, 1.0);
+    //o.position = float4(vertex.pos * float2(1.0 / 1920.0, 1.0 / 1080.0) + float2(0, 0), 0, 1);
```
Jasper-Bekkers, line 28: Nit: commented code.

examples/vulkan-visualization/src/main.rs
```diff
@@ -0,0 +1,1417 @@
...
+            let vs = include_str!("./shaders/imgui.vs.hlsl");
+            let vs = hassle_rs::compile_hlsl(
+                "imgui.vs.hlsl",
+                vs,
+                "main",
+                "vs_5_0",
+                &["-WX", "-spirv", "-fvk-use-scalar-layout"],
+                &[],
+            )
+            .unwrap();
```
Jasper-Bekkers, line 137: It might be nicer to check in pre-compiled .spv binaries along with the shaders and just have a script that calls into dxc instead of using hassle. It's not expected that everybody has a copy of dxcompiler.dll on their system (that can compile SPIR-V code).

Jasper-Bekkers (changes requested):
Good work! Good to see the visualizer out in the open now too. I think it's good to also add some screenshots of it to the README.md file.

examples/vulkan-visualization/src/main.rs
max-traverse: I have moved the imgui renderer to a seperate file. I did the same for one of the helper functions, as it's shared between the example and the imgui renderer.

PR #9, opened 2020-11-25
Functionality for getting a mapped slice from a SubAllocation

max-traverse:
Added functions for getting a mapped slice to the allocated memory. Additionally the mapped pointer in SubAllocation has been turned into an option of a NonNull pointer to be more Rust idiomatic. Another small change is that the size of the allocation is now also stored in the  object, as it's needed to make the mapped slice possible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -164,7 +164,8 @@ pub struct SubAllocation {
     memory_type_index: usize,
     device_memory: vk::DeviceMemory,
     offset: u64,
-    mapped_ptr: *mut std::ffi::c_void,
+    size: u64,
+    mapped_ptr: Option<std::ptr::NonNull<std::ffi::c_void>>,
 
     name: Option<String>,
     backtrace: Option<String>,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -185,14 +186,53 @@ impl SubAllocation {
...
         self.mapped_ptr
     }
 
+    /// Returns a valid mapped slice if the memory is host visible, otherwise it will return None.
+    /// The pointer already points to the exact memory region of the suballocation, so no offset needs to be applied.
+    pub fn mapped_slice(&self) -> Option<&[u8]> {
+        if let Some(ptr) = self.mapped_ptr() {
+            unsafe {
+                Some(std::slice::from_raw_parts(
+                    ptr.as_ptr() as *mut u8,
```
MarijnS95, line 212:
```suggestion
                    ptr.as_ptr() as *const u8,
```
or
```suggestion
                    ptr.as_ptr() as *const _,
```
Bit odd there's no `.as_ptr()` and `.as_mut_ptr()` :shrug:

MarijnS95 (approved):
Nice, simple change that looks good!

One small thing: should `fn size()` return `usize` instead? Conversely, store `size: usize` in a `SubAllocation` and do away with `as usize` casts in `mappped_slice*`?

src/lib.rs
max-traverse: We decided to use `u64` as it's the actual type that is used by both vulkan and d3d12 for the sizes.

PR #14, opened 2021-01-04
Disable storing stack traces by default.

max-traverse:
Generating a stack trace seems to be taking a very long time (1.5 ms) which is not wanted in real-time applications.
This PR changes the default behaviour to not generate any stack traces.


MarijnS95 (approved):
Aside the time overhead this also causes obscure and unpredictable locking errors on Windows: https://github.com/rust-lang/backtrace-rs/issues/399. Disabling is fine, though we should keep in mind this is necessary to repro that bug.

PR #15, opened 2021-01-11
Use bitflags::contains to check for equivalence

MarijnS95:
Seeing a combination of `==`, `!=`, `is_empty()` and comparison against `::empty()` which can all be replaced with simple calls to `.contains`.  This function checks if _all_ bitflags in the RHS are contained in LHS (as opposed to `.intersects` which only checks if there is any flag in common).

At the same time simplify `memory_types` usage by creating a correctly sized slice once. This should probably also be used in `find_memorytype_index`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -371,12 +371,14 @@ impl MemoryType {
             AllocationType::NonLinear
         };
 
-        let memblock_size =
-            if !(self.memory_properties & vk::MemoryPropertyFlags::HOST_VISIBLE).is_empty() {
-                DEFAULT_HOST_MEMBLOCK_SIZE
-            } else {
-                DEFAULT_DEVICE_MEMBLOCK_SIZE
-            };
+        let memblock_size = if self
+            .memory_properties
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -631,39 +635,34 @@ impl VulkanAllocator {
...
-        for i in 0..mem_props.memory_type_count {
-            let flags = mem_props.memory_types[i as usize].property_flags;
-
-            if (flags & vk::MemoryPropertyFlags::HOST_VISIBLE) != vk::MemoryPropertyFlags::empty()
-                && (flags & vk::MemoryPropertyFlags::HOST_COHERENT)
-                    == vk::MemoryPropertyFlags::empty()
-            {
-                log!(Level::Warn, "There is a memory type that is host visible, but not host coherent. It's time to upgrade our memory allocator to take advantage of this type of memory :)");
-            }
+        let host_visible_not_coherent = memory_types.iter().any(|t| {
```
DBouma, line 642: TIL; `.any` in iterators, nice!

DBouma (approved):

PR #16, opened 2021-03-08
Move find_memorytype_index into VulkanAllocator

MarijnS95:
`memory_types` are already stored with the right length in there, no need to keep track of `vk::PhysicalDeviceMemoryProperties` or repeatedly restrict the size of `memory_prop.memory_types`.

---

Apparently I still had this commit pushed but unsubmitted as PR :grimacing:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -570,25 +570,9 @@ impl MemoryType {
     }
 }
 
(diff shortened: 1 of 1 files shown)


max-traverse (approved):
Looks like a good change!

PR #20, opened 2021-05-05
Implement Send + Sync for MemoryBlock

### LINKED ISSUE #10: VulkanAllocator not movable
Not that I want to move it around in memory, but I'd like to have it as a struct member on a struct that is.

Here's the compiler error I do if I try this (the struct it is a member of is `backend::VulkanBackend`):

```
`*mut c_void` cannot be shared between threads safely
within `Option<gpu_allocator::MemoryBlock>`, the trait `Sync` is not implemented for `*mut c_void`
required because it appears within the type `gpu_allocator::MemoryBlock`
required because it appears within the type `Option<gpu_allocator::MemoryBlock>`
required because of the requirements on the impl of `Sync` for `std::ptr::Unique<Option<gpu_allocator::MemoryBlock>>`
required because it appears within the type `alloc::raw_vec::RawVec<Option<gpu_allocator::MemoryBlock>>`
required because it appears within the type `Vec<Option<gpu_allocator::MemoryBlock>>`
required because it appears within the type `gpu_allocator::MemoryType`
required because of the requirements on the impl of `Sync` for `std::ptr::Unique<gpu_allocator::MemoryType>`
required because it appears within the type `alloc::raw_vec::RawVec<gpu_allocator::MemoryType>`
required because it appears within the type `Vec<gpu_allocator::MemoryType>`
required because it appears within the type `VulkanAllocator`
required because it appears within the type `backend::VulkanBackend`
```

Additionally, I'm wondering if it makes sense to make it internally synchronized like vk-mem's, or if I should just wrap it in a mutex.

DBouma:
An unsafe impl Send+Sync should be ok in this case as the mapped pointer's memory outlives the buffer.
Fixes #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -233,31 +233,17 @@ impl SubAllocation {
     /// Returns a valid mapped slice if the memory is host visible, otherwise it will return None.
     /// The slice already references the exact memory region of the suballocation, so no offset needs to be applied.
     pub fn mapped_slice(&self) -> Option<&[u8]> {
-        if let Some(ptr) = self.mapped_ptr {
-            unsafe {
-                Some(std::slice::from_raw_parts(
-                    ptr.as_ptr() as *const _,
-                    self.size as usize,
-                ))
-            }
-        } else {
-            None
-        }
+        self.mapped_ptr.map(|ptr| unsafe {
+            std::slice::from_raw_parts(ptr.as_ptr() as *const _, self.size as usize)
+        })
     }
 
     /// Returns a valid mapped mutable slice if the memory is host visible, otherwise it will return None.
     /// The slice already references the exact memory region of the suballocation, so no offset needs to be applied.
     pub fn mapped_slice_mut(&mut self) -> Option<&mut [u8]> {
-        if let Some(ptr) = self.mapped_ptr {
-            unsafe {
-                Some(std::slice::from_raw_parts_mut(
-                    ptr.as_ptr() as *mut _,
-                    self.size as usize,
-                ))
-            }
-        } else {
-            None
-        }
+        self.mapped_ptr.map(|ptr| unsafe {
+            std::slice::from_raw_parts_mut(ptr.as_ptr() as *mut _, self.size as usize)
+        })
     }
 
     pub fn is_null(&self) -> bool {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -224,14 +224,16 @@ impl SubAllocation {
 
     /// Returns a valid mapped pointer if the memory is host visible, otherwise it will return None.
     /// The pointer already points to the exact memory region of the suballocation, so no offset needs to be applied.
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -361,6 +361,10 @@ impl MemoryBlock {
     }
 }
 
+// The `mapped_ptr` points towards mapped memory that outlives the buffer.
```
MarijnS95, line 364: `Send`/`Sync` not only regard themselves with lifetimes, but with borrows and _sharing_ in the more general case. I think `Send` is fine since the pointer can happily live on when _transferring ownership_ to another thread. `Sync` is not however, since two threads would be able to manage the same pointer on their own regard.
(However, afaik that requires clone or copy to be duplicated in the first place, my mind is still fuzzy on the exact specifications of these :thinking:)

https://doc.rust-lang.org/nomicon/send-and-sync.html

termhn: `Send` is fine but `Sync` is currently not, due to [this method](https://github.com/Traverse-Research/gpu-allocator/blob/39be2dabe38a9f51858d44eb1dca7d17aca7ff3f/src/lib.rs#L246-L259) being marked as safe. The risk being that if another thread had a `mapped_slice` while we try to `mapped_slice_mut`, we get bad things (mutable aliasing). If we marked the `mut` version as unsafe then we could impl `Sync` I believe.

termhn: Wait actually, I'm wrong. `mapped_slice_mut` requires `&mut self` so that's fine. I think `Sync` is actually ok...

MarijnS95: That's true, the only function on `MemoryBlock` moves it and that would be impossible if it were borrowed by multiple threads at the same time, so `Sync` should be safe too.

`SubAllocation` looks good too and is `Send` already, as long as users don't do anything funky to `mapped_ptr` and operate on slices only - we should make that function `unsafe`, something like:

```rust
/// # Safety
/// Returns a raw pointer, lifetime is not guaranteed beyond this point
```

But the same realistically applies to the `vk::DeviceMemory` handle too, though let's not concern ourselves too much with the GPU timeline :).

@DBouma If you can update the comment to mention these things about thread sharing rather than a pointer outliving a buffer (that's more applicable to `SubAllocation::mapped_ptr()`) we can approve and merge this :)

termhn: > /// # Safety
/// Returns a raw pointer, lifetime is not guaranteed beyond this point

I think it is more

```
/// # Safety
/// Be careful not to mutably alias with this pointer; safety cannot be guaranteed, particularly over multiple threads.
```

MarijnS95: Sure, should have been `lifetime and ownership` :smile:

max-traverse (approved):
Looks good to me! The comment about safety is concise and clear.
The added clippy fixes are also nice :D

PR #26, opened 2021-05-25
cargo: Upgrade winit to 0.25

MarijnS95:


MarijnS95:
> Any changes in this version release that are relevant to us?

There are no breaking API changes as far as I'm aware, just that a lot of ancient dependencies have been bumped that are relevant to reduce duplicate deps in our own projects as well.

Aside from that there must have been a ton of bugfixes, probably some new functions here and there, but nothing that stands out.

max-traverse (approved):

PR #27, opened 2021-05-25
Fix lifetime issues by not calling `.build()` on Vulkan builders

MarijnS95:
`.build()` discards all lifetime/borrow information attached to the builder and should be used sparingly - only when a struct is generally passed by value and does not contain pointer references.

In particular local `&[single_item]` slices get dropped at the end of an expression, well before the constructed object is passed into a Vulkan function.  Fortunately creating slices on a single item is as simple as combining the borrow of that single item with a length of `1` into a fat-pointer through `std::slice::from_ref`.  When multiple items have to be passed `.build()` is best called as late as possible, in the same expression as the Vulkan call receiving the objects.

At the same time `from_ref` helps passing builders into Vulkan functions (and other builders): builders `Deref` into their Vulkan struct natrually through `std::slice::from_ref(&some_builder)`, which does not happen when the builder is moved by-value into a temporary array through `&[some_builder]`.

In places where it is hard or obnoxious to remove `.build()` - and structs do not contain pointers - the builder has been replaced with a direct struct constructor, being more concise while at the same time not demonstrating this "bad" `.build()` pattern.

### DIFF
--- a/examples/vulkan-visualization/src/imgui_renderer.rs
+++ b/examples/vulkan-visualization/src/imgui_renderer.rs
@@ -51,37 +51,37 @@ impl ImGuiRenderer {
     ) -> Result<Self, vk::Result> {
         let (pipeline_layout, descriptor_set_layouts) = {
             let bindings = [
-                vk::DescriptorSetLayoutBinding::builder()
-                    .binding(0)
-                    .descriptor_type(vk::DescriptorType::UNIFORM_BUFFER)
-                    .descriptor_count(1)
-                    .stage_flags(vk::ShaderStageFlags::VERTEX)
-                    .build(),
-                vk::DescriptorSetLayoutBinding::builder()
-                    .binding(1)
-                    .descriptor_type(vk::DescriptorType::SAMPLER)
-                    .descriptor_count(1)
-                    .stage_flags(vk::ShaderStageFlags::FRAGMENT)
-                    .build(),
-                vk::DescriptorSetLayoutBinding::builder()
-                    .binding(2)
-                    .descriptor_type(vk::DescriptorType::SAMPLED_IMAGE)
-                    .descriptor_count(1)
-                    .stage_flags(vk::ShaderStageFlags::FRAGMENT)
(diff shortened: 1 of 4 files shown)


examples/vulkan-visualization/src/imgui_renderer.rs
```diff
@@ -144,18 +140,15 @@ impl ImGuiRenderer {
...
-            ];
+            let vertex_stage = vk::PipelineShaderStageCreateInfo::builder()
+                .stage(vk::ShaderStageFlags::VERTEX)
+                .module(vs_module)
+                .name(std::ffi::CStr::from_bytes_with_nul(b"main\0").unwrap());
+            let fragment_stage = vk::PipelineShaderStageCreateInfo::builder()
+                .stage(vk::ShaderStageFlags::FRAGMENT)
+                .module(ps_module)
+                .name(std::ffi::CStr::from_bytes_with_nul(b"main\0").unwrap());
+            let stages = [vertex_stage.build(), fragment_stage.build()];
```
MarijnS95, line 151: This is the only remaining place with serious badness... But it's okay since the `CStr`s borrow the literal, which is `&'static`.

examples/vulkan-visualization/src/imgui_renderer.rs
```diff
@@ -558,35 +538,43 @@ impl ImGuiRenderer {
...
+            let image_info = vk::DescriptorImageInfo::builder().sampler(sampler);
+            let sampler = vk::WriteDescriptorSet::builder()
+                .dst_set(descriptor_sets[0])
+                .dst_binding(1)
+                .descriptor_type(vk::DescriptorType::SAMPLER)
+                .image_info(std::slice::from_ref(&image_info));
+
+            let image_info = vk::DescriptorImageInfo::builder()
+                .image_view(font_image_view)
+                //.sampler(imgui_renderer.sampler)
```
max-traverse, line 560: I think this can be removed. Not sure why it was in there before.

PR #28, opened 2021-05-25
visualization: Remove channel and only render on MainEventsCleared

MarijnS95:
~~Depends on #27~~

With the loop and channel we were previously redrawing and rendering a frame _for every event_.  With vsync enabled this immediately locks up the application as mouse polling rates are well above that.

We cleaned our own applications to embrace winit's "new" eventloop closure instead of dealing with all the desync, race conditions and other oddities when using a separate thread to handle the events, and do the same here too.

---

Still draft, need to check why we're freeing buffers that are still in active command buffers, and why the application is extremely slow to drag around.

### DIFF
--- a/examples/vulkan-visualization/src/main.rs
+++ b/examples/vulkan-visualization/src/main.rs
@@ -3,10 +3,7 @@ use ash::vk;
 use std::default::Default;
 use std::ffi::CString;
 
-use gpu_allocator::{
-    vulkan::{Allocator, AllocatorCreateDesc},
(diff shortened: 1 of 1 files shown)


MarijnS95:
If anyone wants to review this I strongly encourage to only look at the SECOND commit. The first one is merely juggling indentation around to make the diff easier to read, nothing else.

MarijnS95:
image

FWIW "hide whitespace changes" seems to do a proper job, too :ok_hand:

PR #29, opened 2021-05-25
visualization: Replace custom event conversion with imgui-winit-support

MarijnS95:
Depends on ~~#27 and~~ #28


MarijnS95:
@max-traverse We should apply #29 and this to the DX12 visualizer as well, but since you plan on removing the duplication in #59 anyway we should probably keep these patches for Vulkan only and make sure to use the Vulkan visualizer as a base when unifying.

PR #30, opened 2021-05-26
Refactor of gpu_allocator to allow for support for more graphics APIs

max-traverse:
One of the main features on the todo list for `gpu-allocator` is D3D12 support. Since `gpu-allocator` was originally written for vulkan support, we need to make some changes to support multiple graphics APIs. This PR does all the prep work to add D3D12 support.

One of the big downsides is that it requires changes to the API. This is mainly due to splitting the graphics API agnostic and graphics API specific code.


examples/vulkan-visualization/src/imgui_renderer.rs
```diff
@@ -15,16 +16,16 @@ pub struct ImGuiRenderer {
 
     vb_capacity: u64,
     ib_capacity: u64,
-    vb_allocation: SubAllocation,
-    ib_allocation: SubAllocation,
+    vb_allocation: Allocation,
+    ib_allocation: Allocation,
```
Jasper-Bekkers, line 20: How come the SubAllocation got renamed to Allocation? I think this is API churn that we shouldn't really need, and it's also an incorrect name, since the this data still refers to a sub allocation of either a VkMemory or ID3D12Heap.

max-traverse: The idea was that not every allocation returned by the allocator is a sub-allocation (allocations taking an entire block). Since this PR will require API changes I think it's not a bad idea to make this change now instead of another time.

Jasper-Bekkers (approved):

PR #31, opened 2021-05-26
Pass chunk_id instead of SubAllocation to SubAllocator::free

MarijnS95:
The allocators are responsible for the underlying allocation algorithm but have little to do with move semantics of the allocation itself, this is the responsibility of the backend-specific allcoator "wrapper" like `VulkanAllocator`.

---

@max-traverse following the conclusion we came to today, wrapped in a concise, documented PR/commit (messing up your rebase once again :grinning:).  While writing this commit message it's pretty clear we need to come up with better naming for the allocator "algorithms" versus the backend-specific talk-to-the-api allocators.

### DIFF
--- a/src/dedicated_block_allocator.rs
+++ b/src/dedicated_block_allocator.rs
@@ -1,7 +1,5 @@
 #![deny(unsafe_code, clippy::unwrap_used)]
-use super::{
-    AllocationError, AllocationType, Result, SubAllocation, SubAllocator, SubAllocatorBase,
-};
+use super::{AllocationError, AllocationType, Result, SubAllocator, SubAllocatorBase};
 use log::{log, Level};
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


src/dedicated_block_allocator.rs
```diff
@@ -53,8 +51,8 @@ impl SubAllocator for DedicatedBlockAllocator {
         Ok((0, dummy_id))
     }
 
-    fn free(&mut self, sub_allocation: SubAllocation) -> Result<()> {
-        if sub_allocation.chunk_id != std::num::NonZeroU64::new(1) {
+    fn free(&mut self, chunk_id: std::num::NonZeroU64) -> Result<()> {
+        if Some(chunk_id) != std::num::NonZeroU64::new(1) {
```
MarijnS95, line 55: These always get me. Yes, we're sure that `::new(1)` never returns `None`, but what if it did and the `sub_allocation` argument was `None` too??

At least that can't happen anymore :grimacing:

max-traverse:
This will get solved in the refactor PR that is open

MarijnS95:
@max-traverse Fwiw this is submitted separately to be a concise change documented in history :wink:

PR #33, opened 2021-06-10
Visualizer: Show number of blocks at the top level

hrydgard:
Lets you waste less time expanding Types that don't have anything in them:

image

Alternatively I guess you could entirely hide such Types.

### DIFF
--- a/src/visualizer/mod.rs
+++ b/src/visualizer/mod.rs
@@ -299,7 +299,7 @@ impl AllocatorVisualizer {
                 {
                     ui.indent();
                     for (mem_type_i, mem_type) in alloc.memory_types.iter().enumerate() {
-                        if CollapsingHeader::new(&ImString::new(format!("Type: {}", mem_type_i,)))
+                        if CollapsingHeader::new(&ImString::new(format!("Type: {} ({} blocks)", mem_type_i, mem_type.memory_blocks.len() )))
                             .build(ui)
                         {
                             let mut total_block_size = 0;
--- a/src/visualizer/mod.rs
+++ b/src/visualizer/mod.rs
@@ -299,8 +299,13 @@ impl AllocatorVisualizer {
                 {
                     ui.indent();
                     for (mem_type_i, mem_type) in alloc.memory_types.iter().enumerate() {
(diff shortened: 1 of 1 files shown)


Jasper-Bekkers (approved):

MarijnS95 (approved):
Definitely!

On mobile it looks like there's a trailing space in the call, the CI will probably fail on that in fmt.

max-traverse:
This may conflict with ImGUI's label based state tracking. I suspect it will automatically close the collapsing header when the number of blocks change. I know of using `some_hidden_name##visible name` to avoid ID conflicts, but I don't know on top of my head if there is a way to specify an ID that is not linked to the label.
I'll try to find some time today to double check this, as I do think this is a good feature.

max-traverse (changes requested):
Ah I found the issue with a solution in this github issue: https://github.com/ocornut/imgui/issues/1481
If you could apply the same fix here I think it's ready for merging.

hrydgard:
Not sure exactly which solution in that thread you're referring to? Seems all over the place to me.

hrydgard:
I implemented what looks like the simplest solution - hope that's what you meant.

MarijnS95:
@hrydgard It appears @max-traverse is referring to the tree posts from the author of ImGUI: https://github.com/ocornut/imgui/issues/1481#issuecomment-349718791. Supposedly `TreeNode` should be used over `CollapsingHeader` as the latter one is "not an API designed for that sort of data" (third comment), the second comment shows how to implement that.

hrydgard:
Yeah, in C++.

That's a bit of a bigger change than I have time for, and what I implemented works fine.

MarijnS95:
@hrydgard And it's 2017, too. I cannot even find the bindings for most of the functions, let alone those sys bindings being used by the high-level wrapper. Fine by me to leave it as-is, unless Max sees a different way forward.

max-traverse:
I was talking about the `###`. Could you check if increasing/decreasing the number of blocks doesn't open/close the header?

hrydgard:
Right, but that's not what you linked to.

Either way, I do use ### here and it works as expected.

hrydgard:
Can we get this merged? I've tested, it works.

max-traverse:
It looks good to me! Let's merge it in. Thank you for your contribution ^.^

max-traverse (approved):

PR #34, opened 2021-06-16
Make whether the device supports BDA or not a runtime setting.

hrydgard:
BDA = buffer_device_address

(Should we also remove the cfg! check?)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,6 +14,7 @@
 //!     device,
 //!     physical_device,
 //!     debug_settings: Default::default(),
+//!     device_supports_bda: true,  // Ideally, check the BufferDeviceAddressFeatures struct.
 //! });
 //! ```
 //!
@@ -32,6 +33,7 @@
 //! #     device,
 //! #     physical_device,
 //! #     debug_settings: Default::default(),
+//! #     device_supports_bda: true,  // Ideally, check the BufferDeviceAddressFeatures struct.
 //! # });
 //!
 //! // Setup vulkan info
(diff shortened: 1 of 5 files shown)


Jasper-Bekkers:
We've had the cfg there only for legacy reasons, so it should be fine to remove by now. Is there a way we can just query the feature when creating the allocator easily instead of passing in a bool?

hrydgard:
Yeah, I suppose I can add code to do that, but it feels kinda redundant since a calling application, to be portable, must already check for it.

src/lib.rs
```diff
@@ -135,6 +137,7 @@ pub struct VulkanAllocatorCreateDesc {
     pub device: ash::Device,
     pub physical_device: ash::vk::PhysicalDevice,
     pub debug_settings: AllocatorDebugSettings,
+    pub device_supports_bda: bool,
```
MarijnS95, line 140: Since we're going for the `BDA` abbreviation in a _public struct_ this should - just like the PR description - clarify the meaning of the field with a doc-comment :wink:

MarijnS95: I would have been fine with `device_supports_buffer_device_address` too, or just `buffer_device_address`.

MarijnS95: Then, it is not solely about whether the device supports it but also whether the caller _wants_ it.

hrydgard: I'll make it buffer_device_address.

hrydgard: Done.

MarijnS95 (approved):

hrydgard:
Can we get a merge? (and maybe a bump?) :)

max-traverse:
Looks good! Thanks you @hrydgard

PR #36, opened 2021-07-02
Implement Debug for VulkanAllocator. 

hrydgard:
An easy way to dump out the whole state in a panic.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -611,7 +611,7 @@ impl fmt::Debug for VulkanAllocator {
             .field("memory_types", &self.memory_types)
             .field("buffer_image_granularity", &self.buffer_image_granularity)
             .field("debug_settings", &self.debug_settings)
-            .finish()
+            .finish_non_exhaustive()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,7 @@
 use ash::version::{DeviceV1_0, InstanceV1_0};
 use ash::vk;
 use log::{log, Level};
+use std::fmt;
 
 mod result;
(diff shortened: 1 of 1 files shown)


MarijnS95:
Would it make sense to use `finish_non_exhaustive` to indicate that not all fields are printed, despite them being "internal" or "irrelevant"? I wonder if `memory_heaps` could be interesting to see, even if only used for the visualizer.

hrydgard:
memory_heaps only contains vk::MemoryHeap pointers, not too interesting on its own. Could of course query out information from them I guess, but seems out of scope for an initial take. And doing `#[cfg(feature = "visualizer")]` in the middle of a chain doesn't work (can of course rewrite it so it does work, but it better be worth it and doesn't seem to be).

Switching to finish_non_exhaustive though, didn't know about that one.

MarijnS95 (approved):

MarijnS95:
FWIW `vk::MemoryHeap` aren't pointers but structs (at specific array indices) holding the size of a heap and its flags. It may be interesting in the future, but for now `finish_non_exhaustive` suffices. Thanks!

hrydgard:
Oh, my bad then! Curious why it's only there when the visualizer is enabled?

MarijnS95:
The heaps (or rather, the indices in `MemortyType::heap_index`) are interesting to see what `MemoryType` lives where (ie. if certain types share the same physical memory area) and how large that area of physical memory is, but are not relevant to the allocation algorithm, given that flags such as `DEVICE_LOCAL` and `HOST_VISIBLE` are fully covered in `MemoryType::property_flags`.

PR #37, opened 2021-07-02
More debug info

hrydgard:
Followup to #36 . Also show the vk::MemoryHeap structs.

Removed the visualizer conditional - I think it's not too expensive to just always keep these around.


Jasper-Bekkers (approved):

MarijnS95 (approved):
Works for me :+1: - can probably leave the comment out, `vk::Device` is an opaque handle anyway.

PR #38, opened 2021-07-02
visualizer: Use Ash built-in bitflags string formatting

MarijnS95:
Ash already defines a stringifyer for bitflags (see `const_debugs.rs`) that is autogenerated based on all available flags, not just what happened to be available at the time this visualizer was written.


max-traverse (approved):
Good improvement!

PR #41, opened 2021-08-16
Add d3d12 support to gpu-allocator

max-traverse:
GPU Allocator has worked great for Vulkan so far, so it is only natural to add support for D3D12. We managed to reuse all the sub-allocation logic from the Vulkan allocator. Both Vulkan and D3D12 will be behind feature flags, so the user can decide which API support they want to compile in there.

Todo:
* [ ] ~Clean-up visualizer code.~
* [x] Update README.md and lib.rs documentation.
* [x] Should the D3D12 allocator take shared ownership over the device by using `AddRef` and `Release`?
* [x] Remove todos.
* [x] Replace panics with errors.


src/d3d12/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+mod visualizer;
+#[cfg(feature = "visualizer")]
+pub use visualizer::AllocatorVisualizer;
+
+use super::allocator;
+use super::allocator::AllocationType;
+
+use crate::{AllocationError, AllocatorDebugSettings, MemoryLocation, Result};
+
+///
```
MarijnS95, line 16: Docs?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+#[derive(Clone, Debug)]
+pub struct AllocationCreateDesc<'a> {
+    /// Name of the allocation, for tracking and debugging purposes
+    pub name: &'a str,
+    /// Location where the memory allocation should be stored
+    pub location: MemoryLocation,
+
+    /// Size of allocation, should be queried using `ID3D12Device::GetResourceAllocationInfo`
+    pub size: u64,
+    /// Alignment of allocation, should be queried using `ID3D12Device::GetResourceAllocationInfo`
```
MarijnS95, line 67:
```suggestion
    /// Size of allocation, should be queried using [`d3d12::ID3D12Device::GetResourceAllocationInfo()`]
    pub size: u64,
    /// Alignment of allocation, should be queried using [`d3d12::ID3D12Device::GetResourceAllocationInfo()`]
```
?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+    pub alignment: u64,
+    /// Resource category based on resource dimension and flags. Can be created from a `D3D12_RESOURCE_DESC`
```
MarijnS95, line 69:
```suggestion
    /// Resource category based on resource dimension and flags. Can be created from a [`d3d12::D3D12_RESOURCE_DESC`]
```

examples/d3d12-visualization/src/main.rs
```diff
@@ -0,0 +1,556 @@
...
+    resource: *mut ID3D12Resource,
+    rtv_handle: D3D12_CPU_DESCRIPTOR_HANDLE,
+}
+
+fn find_hardware_adapter(
+    dxgi_factory: &all_dxgi::IDXGIFactory6,
+) -> Option<*mut all_dxgi::IDXGIAdapter4> {
+    let mut adapter: *mut all_dxgi::IDXGIAdapter4 = std::ptr::null_mut();
+    let mut adapater_index = 0;
+    loop {
```
MarijnS95, line 40:
```suggestion
    for adapter_index in 0..
```

Drop the `+=` in multiple places below.

examples/d3d12-buffer/src/main.rs
```diff
@@ -0,0 +1,281 @@
+use winapi::shared::{dxgiformat, winerror};
+use winapi::um::{d3d12, d3dcommon};
+
+#[cfg(windows)]
+mod all_dxgi {
+    pub use winapi::shared::{dxgi1_3::*, dxgi1_6::*, dxgitype::*};
+}
+
+use log::*;
```
MarijnS95, line 9: This example seems to use a mixture of `error!(...)`, `log!(Type, ...)` and `println!(...)`, can we clean that up?

examples/d3d12-visualization/src/main.rs
```diff
@@ -0,0 +1,556 @@
...
+                std::ptr::null_mut(),
+                &ID3D12GraphicsCommandList::uuidof(),
+                &mut command_list as *mut _ as *mut _,
+            );
+            if FAILED(hr) {
+                panic!("Failed to create command list.");
+            }
+
+            let command_list = command_list.as_mut().unwrap();
+            command_list
```
MarijnS95, line 353: Drop `let` binding?

examples/d3d12-visualization/src/main.rs
```diff
@@ -0,0 +1,556 @@
...
+                        D3D12_RESOURCE_STATE_PRESENT,
+                    )];
+                    command_list.ResourceBarrier(barriers.len() as u32, barriers.as_ptr());
+                }
+
+                command_list.Close();
+
+                let lists: [*mut ID3D12CommandList; 1] =
+                    [command_list as *mut _ as *mut ID3D12CommandList];
+                queue.ExecuteCommandLists(1, lists.as_ptr());
```
MarijnS95, line 502: Doesn't `&command_list as *const _ as *const *mut ID3D12CommandList` work directly here? Or `&command_list as &*mut ID3D12CommandList as *const _`?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+}
+
+impl From<&d3d12::D3D12_RESOURCE_DESC> for ResourceCategory {
+    fn from(desc: &d3d12::D3D12_RESOURCE_DESC) -> Self {
+        if desc.Dimension == d3d12::D3D12_RESOURCE_DIMENSION_BUFFER {
+            ResourceCategory::Buffer
+        } else if (desc.Flags
+            & (d3d12::D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET
+                | d3d12::D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))
+            != 0
```
MarijnS95, line 49: Didn't these types have `.contains`/`.intersects`?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+        self.chunk_id
+    }
+
+    /// Returns the `d3d12::ID3D12Heap` object that is backing this allocation.
+    /// This heap object can be shared with multiple other allocations and shouldn't be freed (or allocated from)
+    /// without this library, because that will lead to undefined behavior.
+    ///
+    /// # Safety
+    /// The result of this function can safely be used to pass into `CreatePlacedResource`. It's exposed
+    /// for this reason. Keep in mind to also pass `Self::offset()` along to it.
```
MarijnS95, line 129: Could use a bunch of `[]` intradoc-links, pointing users to the right bits in the winapi documentation.

src/d3d12/mod.rs
```diff
@@ -0,0 +1,657 @@
...
+    /// Alignment of allocation, should be queried using `ID3D12Device::GetResourceAllocationInfo`
+    pub alignment: u64,
+    /// Resource category based on resource dimension and flags. Can be created from a `D3D12_RESOURCE_DESC`
+    /// using the helper into function. The resource category is ignored when Resource Heap Tier 2 or higher
+    /// is supported.
+    pub resource_category: ResourceCategory,
+}
+
+impl<'a> AllocationCreateDesc<'a> {
+    #[cfg(feature = "winapi")]
```
MarijnS95, line 76: Interesting, and good choice if `windows-rs` becomes a thing with a sharable public API at some point; but for now moot since the below uses `winapi` a whole lot without being limited to this feature toggle.

src/d3d12/mod.rs
```diff
@@ -0,0 +1,650 @@
...
+                //TODO(max): Return error
+                hr,
+                winerror::S_OK,
+                "Failed to allocate ID3D12Heap of {} bytes",
+                size,
+            );
+
+            //TODO(max): What type of error should this be? It's more like an OOM error
+            std::ptr::NonNull::new(heap)
+                .ok_or_else(|| AllocationError::Internal("Failed to create ID3D12Heap".into()))?
```
MarijnS95, line 206: Might as well add `AllocationError::OOM`?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,650 @@
...
+        unsafe { self.heap.as_ref().Release() };
+    }
+}
+
+// `mapped_ptr` is safe to send or share across threads because
+// it is never exposed publicly through [`MemoryBlock`].
+unsafe impl Send for MemoryBlock {}
+unsafe impl Sync for MemoryBlock {}
+
+#[cfg(windows)]
```
MarijnS95, line 233: This and the other two occurrences in this file are already implied by `mod d3d12` being loaded in the first place?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,650 @@
...
+                true,
+            )?;
+
+            let mut block_index = None;
+            for (i, block) in self.memory_blocks.iter().enumerate() {
+                if block.is_none() {
+                    block_index = Some(i);
+                    break;
+                }
+            }
```
MarijnS95, line 280: `find`?

MarijnS95: I think there exists a `let block_index = self.memory_blocks.iter().position(|block| block.is_none());` for exactly this purpose, too.

src/d3d12/mod.rs
MarijnS95: https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-createheap#return-value

Maybe only do this if the resultcode is `E_OUTOFMEMORY`?

src/d3d12/mod.rs
max-traverse: These flags don't as this is a u32 type, but the DXGI flags seem to support it.

max-traverse: oh nevermind, my rust-analyzer is being slow, DXGI doesn't support it either.

src/d3d12/mod.rs
```diff
@@ -0,0 +1,650 @@
...
+        memory_type.allocate(unsafe { self.device.as_mut() }, desc, backtrace.as_deref())
+    }
+
+    pub fn free(&mut self, allocation: Allocation) -> Result<()> {
+        if self.debug_settings.log_frees {
+            let name = allocation.name.as_deref().unwrap_or("<null>");
+            log!(Level::Debug, "Freeing `{}`.", name);
+            if self.debug_settings.log_stack_traces {
+                let backtrace = format!("{:?}", backtrace::Backtrace::new());
+                log!(Level::Debug, "Free stack trace: {}", backtrace);
```
MarijnS95, line 607: Just use `{:?}` in the `log!()`?

src/d3d12/mod.rs
max-traverse: I handle `E_OUTOFMEMORY` as an `OutOfMemory` error. All other errors are now internal errors (including the pointer being null, which shouldn't happen if the `HRESULT` was `S_OK`.

src/d3d12/mod.rs
```diff
@@ -193,17 +193,18 @@ impl MemoryBlock {
...
-                hr,
-                winerror::S_OK,
-                "Failed to allocate ID3D12Heap of {} bytes",
-                size,
-            );
+            if hr == winerror::E_OUTOFMEMORY {
+                return Err(AllocationError::OutOfMemory);
+            } else if hr != winerror::S_OK {
+                return Err(AllocationError::Internal(format!(
+                    "ID3D12Device::CreateHeap failed with hr 0x{:x}",
```
MarijnS95, line 200:
```suggestion
                    "ID3D12Device::CreateHeap failed with hr {:#x}",
```
?

max-traverse: Ah I didn't know about that one.

examples/d3d12-buffer/src/main.rs
max-traverse: I made everything use `error!` and `info!` in this example.

src/d3d12/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+    }
+
+    /// Returns the [`d3d12::ID3D12Heap`] object that is backing this allocation.
+    /// This heap object can be shared with multiple other allocations and shouldn't be freed (or allocated from)
+    /// without this library, because that will lead to undefined behavior.
+    ///
+    /// # Safety
+    /// The result of this function can safely be used to pass into [`CreatePlacedResource()`]. It's exposed
+    /// for this reason. Keep in mind to also pass [`Self::offset()`] along to it.
+    pub unsafe fn heap(&self) -> *mut d3d12::ID3D12Heap {
```
MarijnS95, line 130: Does this need to be unsafe, per the guidelines we set up prior?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+}
+impl MemoryBlock {
+    fn new(
+        device: &mut d3d12::ID3D12Device,
+        size: u64,
+        heap_properties: &d3d12::D3D12_HEAP_PROPERTIES,
+        heap_category: HeapCategory,
+        dedicated: bool,
+    ) -> Result<Self> {
+        let heap = unsafe {
```
MarijnS95, line 177: Only calling `CreateHeap` is unsafe here (unless I missed something. Let's try to minimize the code surface inside `unsafe {}`.

src/d3d12/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+                HeapCategory::All => d3d12::D3D12_HEAP_FLAG_NONE,
+                HeapCategory::Buffer => d3d12::D3D12_HEAP_FLAG_ALLOW_ONLY_BUFFERS,
+                HeapCategory::RtvDsvTexture => d3d12::D3D12_HEAP_FLAG_ALLOW_ONLY_RT_DS_TEXTURES,
+                HeapCategory::OtherTexture => d3d12::D3D12_HEAP_FLAG_ALLOW_ONLY_NON_RT_DS_TEXTURES,
+            };
+
+            let mut heap: *mut d3d12::ID3D12Heap = std::ptr::null_mut();
+
+            let hr =
+                device.CreateHeap(&desc, &d3d12::IID_ID3D12Heap, &mut heap as *mut _ as *mut _);
```
MarijnS95, line 194: In Breda we migrated to:

```suggestion
            let mut heap = std::ptr::null_mut();

            let hr =
                device.CreateHeap(&desc, &d3d12::IID_ID3D12Heap, &mut heap);
            let heap = : heap.cast::<d3d12::ID3D12Heap>();
```

Thoughts?

src/d3d12/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+                            memory_type_index: self.memory_type_index as usize,
+                            heap: mem_block.heap.as_ptr(),
+                            name: Some(desc.name.to_owned()),
+                            backtrace: backtrace.map(|s| s.to_owned()),
+                        });
+                    }
+                    Err(err) => match err {
+                        AllocationError::OutOfMemory => {} // Block is full, continue search.
+                        _ => return Err(err),              // Unhandled error, return.
+                    },
```
MarijnS95: FWIW:
```suggestion
                    Err(AllocationError::OutOfMemory) => {} // Block is full, continue search.
                    Err(err) => return Err(err),            // Unhandled error, return.
```

src/d3d12/mod.rs
```diff
@@ -0,0 +1,637 @@
...
+            Ok(value) => value,
+            Err(err) => match err {
+                AllocationError::OutOfMemory => {
+                    return Err(AllocationError::Internal(
+                        "Allocation that must succeed failed. This is a bug in the allocator."
+                            .into(),
+                    ))
+                }
+                _ => return Err(err),
+            },
```
MarijnS95, line 378: Nested `match` block can be collapsed too, just like above.

MarijnS95 (changes requested):

src/vulkan/mod.rs
```diff
@@ -8,7 +8,7 @@ pub use visualizer::AllocatorVisualizer;
 use super::allocator;
 use super::allocator::AllocationType;
 use ash::vk;
-use log::{log, Level};
+use log::{debug, warn, Level};
```
MarijnS95, line 11: I pushed "global" changes to Vulkan to this branch as well; shall I PR them separately later?

max-traverse: It's probably best to PR this separately. This PR is already huge, so let's try to not let it get bigger.

MarijnS95: Should all be pushed off in the last force-push, let's get this done :)

src/d3d12/mod.rs
max-traverse: I made it unsafe since the vulkan equivalent is also unsafe in this library. If it should be safe I think that should happen in a different PR.

examples/d3d12-visualization/src/main.rs
max-traverse: I have actually come to prefer the array method (although the hardcoded 1 should be replaced). It makes it simple to add more elements later. In my mind, it also represents more how the API is supposed to be used. You pass an array of things by pointer, even though that can be achieved by taking the address of the one element you have.

MarijnS95: > although the hardcoded 1 should be replaced

Yeah that's my main gripe otherwise; looks like there's one case remaining where `.len()` should be used. You're free to remve `[*mut ID3D12CommandList; 1]` though, the cast already fully defines the target type and the size is implicit.

src/d3d12/mod.rs
Jasper-Bekkers: Worst case we can just pass around a `*const ()` instead - loses type safety but in return we get interoperability with both crates.

MarijnS95: Fine to stick with winapi for now and drop this `#[cfg()]` (otherwise you'd have to spread around at least a couple dozen since it's used _everywhere_ in the public API). We'll see later if we:

- Convert the internals to windows-rs for simpler, safer, and more convenient code;
- Stick to winapi ptrs to convert to/from;
- Have a mode (feature) where our API just exposes (`type`def'd) raw pointers and not have any winapi dependency at all.

Until windows-rs comes with a shareable crate that exposes d3d12 types for interoperability between libraries.

MarijnS95 (approved):
Apart one stray `#[cfg(feature = "winapi")]` (see the remaining open comment), this seems okay to get in right now!

src/d3d12/mod.rs
max-traverse: Yeah I think for now using `winapi` should be fine. Once `windows-rs` is in a stable state we can look into switching to that.

MarijnS95 (approved):

max-traverse:
Stray `#[cfg(feature = "winapi")]` has been removed and my checklist has been finished except for restructuring the visualizer a bit. I'll do that in another PR so we can get this huge PR merged.
Now just waiting for CI before merging.

PR #42, opened 2021-08-16
Run the GitHub Actions CI on multiple platforms, with distinct features

MarijnS95:


max-traverse (approved):
Looks like a good solution to restrict the D3D12 examples to only build on windows for the d3d12-support branch.

PR #43, opened 2021-08-23
Replace log! with debug!/warn! macros

MarijnS95:


max-traverse (approved):
It feels like springtime with all these clean-up PRs you are making ^.^

MarijnS95:
> It feels like springtime

It's not supposed to be, but this way we're all set for the next crate release :grin:

PR #45, opened 2021-08-23
Remove winapi from public interface.

max-traverse:


examples/d3d12-buffer/src/main.rs
```diff
@@ -147,7 +147,7 @@ fn main() {
         let mut resource: *mut d3d12::ID3D12Resource = std::ptr::null_mut();
         let hr = unsafe {
             device.as_ref().unwrap().CreatePlacedResource(
-                allocation.heap(),
+                allocation.heap().0 as *mut _,
```
MarijnS95, line 150: How about an `.as_winapi()` that's only available if `winapi-helper` is set?

Cargo.toml
```diff
@@ -48,17 +48,18 @@ required-features = ["vulkan", "visualizer"]
...
 path = "examples/d3d12-visualization/src/main.rs"
-required-features = ["d3d12", "visualizer"]
+required-features = ["d3d12", "winapi-helper", "visualizer"]
 
 
 [features]
 visualizer = ["imgui"]
 vulkan = ["ash"]
 d3d12 = ["winapi"]
+winapi-helper = ["d3d12"]
```
MarijnS95, line 63: Not sure if "helper" is the right word; and this should enable the winapi _and_ d3d12 feature IMO - or just the winapi feature.

Perhaps `public-winapi` is more descriptive?

examples/d3d12-buffer/src/main.rs
max-traverse: I already have a `.to_winapi()`, but it's not public-facing. I would need to optionally make it public (not sure how that works without duplicating).

I never quite understood the difference between `as_` and `to_` in Rust. When should I use one over the other?

max-traverse: I found the difference between `as_` and `to_`. Quite intuitive actually. https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

src/d3d12/mod.rs
```diff
@@ -3,6 +3,43 @@ use log::{log, Level};
...
+pub trait AbstractWinapiPtr<T> {
+    #[allow(clippy::mut_from_ref)]
+    fn as_winapi(&self) -> &mut T;
+}
+
+#[cfg(not(feature = "public-winapi"))]
+trait AbstractWinapiPtr<T> {
+    #[allow(clippy::mut_from_ref)]
+    fn as_winapi(&self) -> &mut T;
+}
```
max-traverse, line 16: I am not happy with this solution, but I am not sure if we have a better alternative.
The duplicate code is not nice, and disabling clippy is almost always a bad idea.

max-traverse: Perhaps something like this would work?
```suggestion
mod need_a_better_name {
    pub trait AbstractWinapiPtr<T> {
        #[allow(clippy::mut_from_ref)]
        fn as_winapi(&self) -> &mut T;
    }
}
#[cfg(feature = "public-winapi")]
pub use need_a_better_name::*;

#[cfg(not(feature = "public-winapi"))]
use need_a_better_name::*;
```

MarijnS95: Yep, great hack, blame Rust for this.

src/d3d12/mod.rs
```diff
@@ -3,6 +3,43 @@ use log::{debug, Level};
...
+#[derive(Debug)]
+pub struct Dx12DevicePtr(pub *mut std::ffi::c_void);
+impl AbstractWinapiPtr<d3d12::ID3D12Device> for Dx12DevicePtr {
+    fn as_winapi(&self) -> &mut d3d12::ID3D12Device {
+        let device = self.0 as *mut d3d12::ID3D12Device;
+        if let Some(device) = unsafe { device.as_mut() } {
+            device
+        } else {
+            panic!("Attempting to cast device null pointer to reference.")
+        }
```
MarijnS95, line 27: That's just `.unwrap("Attempting to cast device null pointer to reference.")`, but in 5 lines.

src/d3d12/mod.rs
```diff
@@ -3,6 +3,43 @@ use log::{debug, Level};
 use winapi::shared::winerror;
 use winapi::um::d3d12;
 
+mod cond_pub_mod {
+    pub trait AbstractWinapiPtr<T> {
+        #[allow(clippy::mut_from_ref)]
```
MarijnS95, line 8: Afaik all winapi functions take self immutable, so why is this returning a mutable reference?

max-traverse: That is true for cases like: ```device.CallSomeFunction()```.
However, when you have a function like this:
```rust
device.CreatePlacedResource(my_heap.as_winapi(), /* ... */);
```
the heap parameter is a `*mut ID3D12Heap`. At that point, it's just simpler to always return a mutable reference, as otherwise, you would need to cast it with `as *const _ as *mut _`.

src/d3d12/mod.rs
max-traverse: You can't pass a string to unwrap? https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap

I did this annoying if let Some so clippy wouldn't complain about using unwrap.
Perhaps these pointer wrappers should use `NonNull`? That way we catch the error sooner?

MarijnS95: Sorry, I meant `.expect()`.

src/d3d12/mod.rs
MarijnS95: That's bad, but for consistency and the usual Rust-correctness we should just have an `.as_winapi()` and `.as_winap_mut()` don't you think? That however requires `my_heap` to be `mut` which should also be fine?

max-traverse: I think having `.as_winapi()` and `.as_winapi_mut()` is the best solution ^.^

src/d3d12/mod.rs
max-traverse: Updated with the usage of expect :D

PR #46, opened 2021-08-23
Fix broken intradoc links and test rustdoc in CI

MarijnS95:
Let us make sure our public crates have sane docs over at https://docs.rs/gpu-allocator/ :)

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -60,19 +60,19 @@ impl Allocation {
         self.chunk_id
     }
 
-    /// Returns the `vk::DeviceMemory` object that is backing this allocation.
+    /// Returns the [`vk::DeviceMemory`] object that is backing this allocation.
     /// This memory object can be shared with multiple other allocations and shouldn't be freed (or allocated from)
     /// without this library, because that will lead to undefined behavior.
     ///
     /// # Safety
-    /// The result of this function can safely be used to pass into `bind_buffer_memory` (`vkBindBufferMemory`),
-    /// `bind_texture_memory` (`vkBindTextureMemory`) etc. It's exposed for this reason. Keep in mind to also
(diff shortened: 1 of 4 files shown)


src/vulkan/mod.rs
```diff
@@ -60,19 +60,19 @@ impl Allocation {
...
     }
 
-    /// Returns the `vk::DeviceMemory` object that is backing this allocation.
+    /// Returns the [`vk::DeviceMemory`] object that is backing this allocation.
     /// This memory object can be shared with multiple other allocations and shouldn't be freed (or allocated from)
     /// without this library, because that will lead to undefined behavior.
     ///
     /// # Safety
-    /// The result of this function can safely be used to pass into `bind_buffer_memory` (`vkBindBufferMemory`),
-    /// `bind_texture_memory` (`vkBindTextureMemory`) etc. It's exposed for this reason. Keep in mind to also
```
MarijnS95, line 69: @max-traverse I suppose you meant `bind_image_memory` here? Also, there's a `2` variant of both functions, but I don't think we should bother to list them here (unless wanting to write a nice summation).

MarijnS95: Also, `# Safety` should clarify what safety constraints the caller should be held accountable for which is not explained in this block.

max-traverse: oh `bind_texture_memory` should 100% be `bind_image_memory`. Not sure how this ended up in the safety section.

max-traverse (approved):
Nice clean-up! Just need to figure out what to do with those bind comments in the safety section.

MarijnS95:
> Just need to figure out what to do with those bind comments in the safety section.

Let's merge this, I'll let you take care of some doc rewriting in a new PR ;)

PR #54, opened 2021-09-02
Disabling buffer_device_address in vulkan examples

EmilioLaiso:
The vulkan examples are not doing the usual checking and enabling of features, so the feature should be disabled by default.
We could do the feature checking and enabling of the feature accordingly, but I didn't want to unnecessarily complicate the example.

Thoughts?


max-traverse:
Disabling sounds like a good idea so the samples can run on more platforms. I don't think we even need the comments there.

Jasper-Bekkers (approved):

max-traverse:
I went ahead and removed the other one. Will merge as soon as the CI is finished.

PR #60, opened 2021-09-07
Ensure allocation sizes do not exceed the heap size for vulkan validation.

janie177:
Inside the vulkan allocator, no allocation attempts will be made when the requested memory size exceeds the size of the heap for that memory type. This prevents vulkan validation layer warnings.

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -608,7 +608,7 @@ impl Allocator {
             None => return Err(AllocationError::NoCompatibleMemoryTypeFound),
         };
 
-        //Do not try create a block if the heap is smaller than the required size (avoids validation warnings).
+        //Do not try to create a block if the heap is smaller than the required size (avoids validation warnings).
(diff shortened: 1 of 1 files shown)


src/vulkan/mod.rs
```diff
@@ -608,12 +608,18 @@ impl Allocator {
             None => return Err(AllocationError::NoCompatibleMemoryTypeFound),
         };
 
-        let allocation = self.memory_types[memory_type_index].allocate(
-            &self.device,
-            desc,
-            self.buffer_image_granularity,
-            backtrace.as_deref(),
-        );
+        //Do not try create a block if the heap is smaller than the required size (avoids validation warnings).
```
MarijnS95, line 611:
```suggestion
        //Do not try to create a block if the heap is smaller than the required size (avoids validation warnings).
```

MarijnS95 (approved):

max-traverse (approved):

PR #61, opened 2021-09-07
Only render after `MainEventsCleared` event

janie177:
In the vulkan visualizer, events were not processed correctly.
Now it waits for the MainEventsCleared to break out of the loop.


max-traverse (approved):
Yay let's not wait for events if we can render :D

MarijnS95:
Neat, that's the important part extracted from https://github.com/Traverse-Research/gpu-allocator/pull/28!

(Which, by the way, we should finish and merge at some point too)

MarijnS95:
Being a bit more descriptive, I updated the title to reflect that this doesn't necessary `break` as that's an implementation detail, but rather that it only allows rendering to happen after this event instead of after every event.

PR #56, opened 2021-09-03
ci: Enable `public-winapi` feature when compiling on Windows

MarijnS95:
Fixes #55

PR #63, opened 2021-09-10
examples/d3d12-visualizer: Fix clippy::erasing_op by copying macro

MarijnS95:
Commit 0d4aa76 ("Replace usage of helper function that is not available in latest released winapi. (#52)") imported the contents of a helper function (macro) that is not available in stable winapi, but leaves a clippy warning in place.  Instead, copy the original macro in place making it easier on the reader of this example to find and relate the all the math ops that are going on here, and fix the clippy warning at the same time.

This lint issue was found in: https://github.com/Traverse-Research/gpu-allocator/pull/56


max-traverse (approved):
Looks good to me

Jasper-Bekkers (approved):

PR #64, opened 2021-09-10
lib: Fix mismatched type mutability in documentation example

MarijnS95:
---- src\lib.rs - (line 93) stdout ----
    error[E0308]: mismatched types
    Error:   --> src\lib.rs:103:27
       |
    12 |     device: Dx12DevicePtr(device as *const _),
       |                           ^^^^^^^^^^^^^^^^^^ types differ in mutability
       |
       = note: expected raw pointer `*mut c_void`
                  found raw pointer `*const _`


max-traverse (approved):
Good catch!

PR #65, opened 2021-09-19
Update imgui to 0.8 and winit to 0.25

MarijnS95:
Depends on #28 and #29

#29 went down on winit version to accomodate `imgui-winit-support 0.7`'s usage of `winit 0.24`.

This also omits some id (label with `##`) assignments which are unnecessary when the item resides inside a stack with a unique item: https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-have-multiple-widgets-with-the-same-label
These labels should only be used when:
- The label changes based on some dynamic variable like a counter while
  still representing the same object;
- Duplicate labels inside the same stack-item.


max-traverse (approved):
Just need to make it pass CI and then it's good to merge 👍

PR #68, opened 2021-09-20
First go at implementing debug for d3d12 Allocator.

max-traverse:
The `D3D12_HEAP_PROPERTIES` struct does not implement `Debug`, so we can't do a simple `[derive(Debug)]`. Since I don't really know what I am doing when it comes to implementing this manually, I just did what I found to be the simplest for debugging output. Feedback is welcome ^.^


src/d3d12/mod.rs
```diff
@@ -275,6 +276,35 @@ struct MemoryType {
...
+                &self.heap_properties.MemoryPoolPreference,
+            )
+            .field(
+                "heap_properties.CreationNodeMask",
+                &self.heap_properties.CreationNodeMask,
+            )
+            .field(
+                "heap_properties.VisibleNodeMask",
+                &self.heap_properties.VisibleNodeMask,
+            )
```
MarijnS95, line 301: The only proper way I'm aware of to create debuggers for nested structs that do not implement debug, and get all the pretty features from `debug_struct`, is by simply creating a (private) wrapper (new-)type that you define debug on:

```suggestion
struct HeapPropertiesDebug(d3d12::D3D12_HEAP_PROPERTIES);

impl std::fmt::Debug for HeapPropertiesDebug {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("D3D12_HEAP_PROPERTIES")
            .field("Type", &self.0.Type)
            .field("CPUPageProperty", &self.0.CPUPageProperty)
            .field("MemoryPoolPreference", &self.0.MemoryPoolPreference)
            .field("CreationNodeMask", &self.0.CreationNodeMask)
            .field("VisibleNodeMask", &self.0.VisibleNodeMask)
            .finish()
    }
}

impl std::fmt::Debug for MemoryType {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        f.debug_struct("MemoryType")
            .field("memory_blocks", &self.memory_blocks)
            .field("memory_location", &self.memory_location)
            .field("heap_category", &self.heap_category)
            .field(
                "heap_properties",
                &HeapPropertiesDebug(self.heap_properties),
            )
```

(This crates a copy, you might want to sprinkle some borrows and lifetimes around too)

PR #69, opened 2021-09-24
remove macos and ios limitation

jhvst:
Hi there,

Since the 0.9.0 upgrade, the use of the library is broken at least on Apple M1. It seems that a Vulkan limitation was erroneously added to these platforms. I am running my project on this branch https://github.com/periferia-labs/rivi-loader/tree/ramp using changes of this fork, and all is good again.

To my understanding, the moltenvk support is already transparent. You can see from my project that moltenvk does not need to be added to Cargo.toml either for Vulkan to work.

I am happy to give more information about my current environment if you have disparities.

Another thing is that I cannot speak for iOS nor do I have x86 mac available this week. For M1 (target_arch = "aarch64") it works fine.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -158,7 +158,6 @@ pub(crate) mod allocator;
 #[cfg(feature = "visualizer")]
 pub mod visualizer;
 
-#[cfg(all(not(any(target_os = "macos", target_os = "ios")), feature = "vulkan"))]
 pub mod vulkan;
 
 #[cfg(all(windows, feature = "d3d12"))]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -158,6 +158,7 @@ pub(crate) mod allocator;
 #[cfg(feature = "visualizer")]
 pub mod visualizer;
 
+#[cfg(feature = "vulkan")]
 pub mod vulkan;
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -158,7 +158,6 @@ pub(crate) mod allocator;
 #[cfg(feature = "visualizer")]
 pub mod visualizer;
 
-#[cfg(all(not(any(target_os = "macos", target_os = "ios")), feature = "vulkan"))]
```
MarijnS95, line 161: You should still leave the `vulkan` feature in place:

```suggestion
#[cfg(feature = "vulkan")]
```

Other than that I agree that `ash` with `ash-molten` should work just fine on Apple products.

MarijnS95 (changes requested):

jhvst:
seems like i have no idea how github works nowadays, i made the asked changes in a separate commit.

anyway, thanks! maybe you want to add a CI pipeline for M1 but otherwise this seems good to me 👍

max-traverse (approved):
Ah I didn't consider the moltenVK use case. Sounds like a good idea to let the user of the library decide whether to compile in vulkan or not.

MarijnS95 (approved):

PR #70, opened 2021-09-26
reduce size of Allocation structs

udoprog:
This changes the name and backtraces fields on `Allocation` to be `Box<str>` instead of `String`, netting us a `mem::size_of::<Allocation>()` decrease from 104 to 88 bytes.

Same could be done on `AllocationError`, but would be an API break.


MarijnS95 (approved):
Nice to save 8 bytes per-`String` on the `capcity` field, these are "immutable" anyway :)

max-traverse (approved):
This looks like it makes sense for `gpu-allocator`. Thank you very much for your contribution! ^.^

PR #72, opened 2021-09-27
pack memory bits to reduce the size of Allocation on Vulkan

udoprog:
Follow up on #70, this saves an additional 8 bytes on my 64-bit system (presumably due to improved struct packing since the change ~~only saves 6~~ 12).

See the documentation on the `memory_bits` field for rationale on why I think this is sufficient. But these assumptions might change if ever block size implementations change.

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -5,6 +5,9 @@ mod visualizer;
 #[cfg(feature = "visualizer")]
 pub use visualizer::AllocatorVisualizer;
 
+#[cfg(test)]
+mod tests;
+
 use super::allocator;
 use super::allocator::AllocationType;
 use ash::vk;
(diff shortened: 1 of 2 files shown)


MarijnS95:
> this saves an additional 8 bytes on my 64-bit system (presumably due to improved struct packing since the change only saves 6).

If I'm not mistaken `usize` is 8 bytes on 64-bit leading to a 16-byte removal to be replaced with 4 bytes of the `u32`, netting 12 bytes. It does not appear that any better struct packing is possible since every other field is 8 bytes in size and probably mandates alignment of the same. In turn we have four more bytes of headroom before struct size increases :)

udoprog:
Yeah. Was working on a different thing at the same time and got mixed up. Cheers!

max-traverse:
Just wondering if you have a specific use case for this optimization. It seems a bit overkill to add this complexity to save a couple of bytes per allocation.

Jasper-Bekkers:
@udoprog I'll close this for now due to lack of updates and because it's unclear how much these optimizations will help in practice. If you have a usecase for them, please do let us know and we'll reconsider.

PR #83, opened 2021-11-29
🤷‍♂️ Adding an `Option<&mut bool>` argument in the allocator imgui rendering

EmilioLaiso:
This allows calling code to do the following:

- Pass None to just render the editor as usual
- Pass Some(&mut bool) to control if the window is rendered and get information in return regarding a request for closure of the window (user clicks X button on the imgui window).
This allows for easier integration with UI systems.

I had conversations with @MarijnS95 about this and we're still not 100% happy with the code, adding the draft PR because I need to reference it in another project :P

### DIFF
--- a/src/d3d12/visualizer.rs
+++ b/src/d3d12/visualizer.rs
@@ -3,10 +3,7 @@
 use super::Allocator;
 use crate::visualizer::ColorScheme;
 
-use __core::ops::Deref;
 use winapi::um::d3d12::*;
-
-use imgui::*;
 
 // Default value for block visualizer granularity.
 const DEFAULT_BYTES_PER_UNIT: i32 = 1024;
@@ -90,7 +87,6 @@ impl AllocatorVisualizer {
         }
 
         window
-            .collapsed(true, Condition::FirstUseEver)
             .size([512.0, 512.0], imgui::Condition::FirstUseEver)
             .build(ui, || {
                 use imgui::*;
@@ -284,7 +280,7 @@ impl AllocatorVisualizer {
 
     pub fn render(&mut self, allocator: &Allocator, ui: &imgui::Ui, opened: Option<&mut bool>) {
         let draw = if let Some(o) = opened.as_ref() {
-            *o.deref()
+            **o
         } else {
             true
         };
--- a/src/d3d12/visualizer.rs
+++ b/src/d3d12/visualizer.rs
@@ -276,10 +276,10 @@ impl AllocatorVisualizer {
 
     /// Renders imgui widgets.
     /// The `Option<&mut bool>` can be used control and track changes to the opened/closed status of the widget.
-    /// Pass `None` if no control and read-back information is required. This will always render the widget.
+    /// Pass [`None`] if no control and readback information is required. This will always render the widget.
     /// When passing `Some(&mut bool)`:
-    ///  - If the bool is false, the widget won't be drawn.
-    ///  - If the bool is true the widget will be drawn and an (X) closing button will be added to the widget bar.
+    /// - If [`false`], the widget won't be drawn.
+    /// - If [`true`], the widget will be drawn and an (X) closing button will be added to the widget bar.
     pub fn render(&mut self, allocator: &Allocator, ui: &imgui::Ui, opened: Option<&mut bool>) {
         if opened != Some(&mut false) {
             self.render_main_window(ui, opened, allocator);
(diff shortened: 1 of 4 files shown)


src/d3d12/visualizer.rs
```diff
@@ -77,9 +75,18 @@ impl AllocatorVisualizer {
...
+    fn render_main_window(&mut self, ui: &imgui::Ui, opened: Option<&mut bool>, alloc: &Allocator) {
+        let mut window = imgui::Window::new("Allocator visualization");
+
+        if let Some(opened) = opened {
+            if *opened {
+                window = window.opened(opened);
+            } else {
+                return;
+            }
+        }
```
MarijnS95, line 87: For general use, I think the precedent should be that you don't call this function if you don't want the UI to be rendered.

```suggestion
        if let Some(opened) = opened {
            window = window.opened(opened);
        }
```

src/d3d12/visualizer.rs
```diff
@@ -271,8 +278,17 @@ impl AllocatorVisualizer {
...
+            **o
+        } else {
+            true
+        };
+
+        self.render_main_window(ui, opened, allocator);
+
+        if draw {
+            self.render_memory_block_windows(ui, allocator);
+        }
```
MarijnS95, line 292: Then, given the comment above that we shouldn't call `render_main_window` if it's `Some(false)` (ie. the user/caller _explicitly_ - by passing `Some` - asks for this window to _not_ be shown) you can massively simplify this with:

```suggestion
        if opened != Some(&mut false) {
            self.render_main_window(ui, opened, allocator);
            self.render_memory_block_windows(ui, allocator);
        }
```

If the caller doesn't care about showing or hiding the window and passes `None` this code will always run (and as discussed offline, not calling `.opened()` causes the `x` button to remain hidden).

By the way, don't forget to add some doc-comments explaining this use :)

MarijnS95 (changes requested):

src/d3d12/visualizer.rs
```diff
@@ -271,8 +274,16 @@ impl AllocatorVisualizer {
...
 
-    pub fn render(&mut self, allocator: &Allocator, ui: &imgui::Ui) {
-        self.render_main_window(ui, allocator);
-        self.render_memory_block_windows(ui, allocator);
+    /// Renders imgui widgets.
+    /// The `Option<&mut bool>` can be used control and track changes to the opened/closed status of the widget.
+    /// Pass `None` if no control and read-back information is required. This will always render the widget.
+    /// When passing `Some(&mut bool)`:
+    ///  - If the bool is false, the widget won't be drawn.
+    ///  - If the bool is true the widget will be drawn and an (X) closing button will be added to the widget bar.
```
MarijnS95, line 282:
```suggestion
    /// Pass [`None`] if no control and readback information is required. This will always render the widget.
    /// When passing `Some(&mut bool)`:
    /// - If [`false`], the widget won't be drawn.
    /// - If [`true`], the widget will be drawn and an (X) closing button will be added to the widget bar.
```

max-traverse:
@EmilioLaiso Are there still changes you want to make to this since it's a draft, or can we merge this?

MarijnS95:
@EmilioLaiso you can also apply the doc-comment change to the other documentation in this PR :)

src/vulkan/visualizer.rs
```diff
@@ -258,8 +261,17 @@ impl AllocatorVisualizer {
         self.focus = None;
     }
 
-    pub fn render(&mut self, allocator: &Allocator, ui: &imgui::Ui) {
-        self.render_main_window(ui, allocator);
-        self.render_memory_block_windows(ui, allocator);
+    /// Renders imgui widgets.
+    ///
+    /// The `Option<&mut bool>` can be used control and track changes to the opened/closed status of the widget.
```
MarijnS95, line 266:
```suggestion
    /// The [`Option<&mut bool>`] can be used control and track changes to the opened/closed status of the widget.
```

src/d3d12/visualizer.rs
```diff
@@ -271,8 +274,17 @@ impl AllocatorVisualizer {
         self.focus = None;
     }
 
-    pub fn render(&mut self, allocator: &Allocator, ui: &imgui::Ui) {
-        self.render_main_window(ui, allocator);
-        self.render_memory_block_windows(ui, allocator);
+    /// Renders imgui widgets.
+    ///
+    /// The `Option<&mut bool>` can be used control and track changes to the opened/closed status of the widget.
```
MarijnS95, line 279:
```suggestion
    /// The [`Option<&mut bool>`] can be used control and track changes to the opened/closed status of the widget.
```

MarijnS95 (approved):

PR #88, opened 2021-12-22
vulkan: Remove unused `backtrace` field from `Allocator`

MarijnS95:
Only `SubAllocator`s print their backtrace and need to hold on to this field; there is no need for the `Allocator` to hold on to and own the exact same value.

Rust 1.57 now detects such unused fields, when they have previously only been used inside `derive()` macros such as `Debug`.

### DIFF
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -149,7 +149,6 @@ pub struct Allocation {
     heap: *mut d3d12::ID3D12Heap,
 
     name: Option<Box<str>>,
-    backtrace: Option<Box<str>>,
 }
 
 unsafe impl Send for Allocation {}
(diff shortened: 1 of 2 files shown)


Jasper-Bekkers (approved):

Jasper-Bekkers:
:thinking: should we pin the CI builds to a certain rust version and just update that every once in a while to prevent random PRs from breaking going forward?

MarijnS95:
> should we pin the CI builds to a certain rust version and just update that every once in a while to prevent random PRs from breaking going forward?

Perhaps - though I much prefer a CI automatically re-running on a new Rust release to at least let us know of these issues. Otherwise new contributors will clone this repo and be faced a ton of irrelevant lint/check warnings they have nothing to do with, because we forgot to update the pinned rust version for some time.

MarijnS95:
We still have a few more of these issues when building without the visualizer. But because the CI doesn't test that usecase I'll see it through for now.

PR #90, opened 2021-12-27
Upgrade to ash 0.35 and ash-window 0.9

MarijnS95:


Cargo.toml
```diff
@@ -21,18 +21,19 @@ include = [
 backtrace = "0.3"
 log = "0.4"
 thiserror = "1.0"
-# Only needed for vulkan. Disable the default "linked" feature to prevent forcibly
-# dynamic-linking against a Vulkan library on the system.  This linking, and loading
-# an entrypoint for it is only needed by the examples.
-ash = { version = "0.34", optional = true, default-features = false, features = ["debug"] }
+# Only needed for vulkan.  Disable all default features as good practice,
+# such as the ability to link/load a Vulkan library.
+ash = { version = "0.35", optional = true, default-features = false, features = ["debug"] }
```
Jasper-Bekkers, line 26: Should we add a `debug` feature to gpu-allocator too so we can disable it entirely when needed?

MarijnS95: Ideally this gets picked up at the Rust compiler side so that [this recommended implementation](https://rust-lang.github.io/api-guidelines/debuggability.html#all-public-types-implement-debug-c-debug) does not suffer nearly as much of a compile-time hit as it does today (25% of the entirety of the Ash crate!).

We probably can and should hook it behind a debug feature for now, but leave it enabled by default for smoother transitioning in downstream creates, just like we opted for Ash. We'll need to make sure the CI checks with and without (if it doesn't already) so that an unguarded `derive(Debug)` is quickly caught.

Cargo.toml
```diff
@@ -21,31 +21,32 @@ include = [
...
 # Only needed for d3d12.
 winapi = { version = "0.3.9", features = ["d3d12", "winerror", "impl-default", "impl-debug"], optional = true }
 # Only needed for visualizer.
 imgui = { version = "0.8", optional = true }
 imgui-winit-support = { version = "0.8", optional = true, default-features = false, features = ["winit-26"] }
 
 [dev-dependencies]
-ash-window = "0.8"
+# Enable the "loaded" feature to be able to access the Vulkan entrypoint.
+ash = { version = "0.35", default-features = false, features = ["debug", "loaded"] }
```
MarijnS95, line 35: Since I don't think we want to install `libvulkan-dev` on the CI just to lint the examples, we're using runtime loading now. Even though compile-time linking makes more sense for examples that are usually compiled and executed on the same machine, and we already depend on the SDK for the validation layers.

examples/vulkan-visualization/src/main.rs
```diff
@@ -14,7 +14,7 @@ use imgui_renderer::ImGuiRenderer;
 use imgui_winit_support::{HiDpiMode, WinitPlatform};
 
 fn main() -> ash::prelude::VkResult<()> {
-    let entry = ash::Entry::new();
+    let entry = unsafe { ash::Entry::load() }.unwrap();
```
MarijnS95, line 17: :(

Cargo.toml
MarijnS95: If we do, this should be picked up in a separate PR anyway.

PR #91, opened 2021-12-28
Update to windows-rs 0.29 

Jasper-Bekkers:
Also make it into a dev dependency.


Cargo.toml
```diff
@@ -38,16 +38,15 @@ raw-window-handle = "0.4"
 winapi = { version = "0.3.9", features = ["d3d12", "d3d12sdklayers", "dxgi1_6", "winerror", "impl-default", "impl-debug", "winuser", "windowsx", "libloaderapi"] }
 winit = "0.26"
 
-[dependencies.windows]
-version = "0.25"
+[dev-dependencies.windows]
+version = "0.29"
```
MarijnS95, line 42: Good catch! I hope to migrate the core of this crate to `windows-rs` at some point, but not until the relation between `windows` and `windows-sys` has been defined, wrappers (the string helpers) exposing completely UB behaviour behind "safe" functionality revised, and a bit more stability between version upgrades in general (seeing open issues, some significant changes may still pass by first).

Cargo.toml
```diff
@@ -38,16 +38,15 @@ raw-window-handle = "0.4"
...
 features = [
     "Win32_Foundation",
     "Win32_Graphics",
-    "Win32_Graphics_Direct3D11",
+    "Win32_Graphics_Direct3D",
```
MarijnS95, line 46: :tada:

examples/d3d12-buffer-winrs/src/main.rs
```diff
@@ -89,7 +90,7 @@ fn main() -> Result<()> {
     })
     .unwrap();
 
-    // Test allocating GPU Only memory
+    // Test allocating Gpu Only memory
```
MarijnS95, line 93: While fixing these inconsistencies (including `test allocation` rewording below) here in `d3d12-buffer-winrs`, please also fix them in `d3d12-buffer` where they were copied from verbatim.

MarijnS95 (approved):

examples/d3d12-buffer/src/main.rs
```diff
@@ -224,11 +224,11 @@ fn main() {
         info!("Allocation and deallocation of CpuToGpu memory was successful.");
     }
 
-    // Test allocating GPU to CPU memory
+    // Test allocating Gpu to Cpu memory
     {
         let test_buffer_desc = d3d12::D3D12_RESOURCE_DESC {
             Dimension: d3d12::D3D12_RESOURCE_DIMENSION_BUFFER,
-            Alignment: 0, // alias for D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT
+            Alignment: 0, // Alias for D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT
```
MarijnS95, line 231: Also capitalize this in `*-winrs` please :)

Doesn't windows-rs ship with these nowadays?

https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Graphics/Direct3D12/constant.D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT.html - the value is not `0` though...

MarijnS95: Ah, according to https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_heap_desc is defined as 64KiB (as we've seen in windows-rs docs above) and `0` is a literal alias meaning `64KiB` inside the API. Indeed much better to use that constant directly than documenting that it's supposed to be the same :+1:

PR #93, opened 2022-01-04
Depend on Windows crates only on Windows

repi:
Avoids downloading and depending on the Windows crates when building of other platforms, without impacting Windows support. 

Noticed that `gpu-allocator` was the only dependency that we had that brought in `winapi` when building our codebase for Mac and Linux, and including it in our licensing summaries and such. This fixes that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,21 +24,25 @@ thiserror = "1.0"
 # Only needed for vulkan.  Disable all default features as good practice,
 # such as the ability to link/load a Vulkan library.
 ash = { version = "0.35", optional = true, default-features = false, features = ["debug"] }
-# Only needed for d3d12.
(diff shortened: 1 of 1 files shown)


max-traverse (approved):
I didn't even realize we were building `winapi` on Linux and MacOS. It makes a lot more sense to only include it on Windows. Thanks!

MarijnS95:
On Linux winapi has a nice `#![cfg(windows)]` in `lib.rs` so it becomes a no-op, but this is even better :)

repi:
Thanks for fast review and merge!

Would it be possible to do a patch release of the crate soon that includes this?

max-traverse:
> Thanks for fast review and merge!
> 
> Would it be possible to do a patch release of the crate soon that includes this?

@Jasper-Bekkers Can you do a new `gpu-allocator` release?

PR #94, opened 2022-01-06
Move imgui-winit-support to dev-dependencies.

max-traverse:
For some reason `imgui-winit-support` was part of the regular dependencies when the `visualizer` was enabled. However, this crate is only used in the example projects, so it should be part of the `dev-dependencies`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ thiserror = "1.0"
 ash = { version = "0.35", optional = true, default-features = false, features = ["debug"] }
 # Only needed for visualizer.
 imgui = { version = "0.8", optional = true }
-imgui-winit-support = { version = "0.8", optional = true, default-features = false, features = ["winit-26"] }
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
Nice!

MarijnS95:
While at it, we should mention the context for this change here in the PR and commit so that others can easily find it. We were running into:

```rust
error: Please select a feature to build for unix: `x11`, `wayland`
  --> .cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.26.0/src/platform_impl/linux/mod.rs:10:1
   |
10 | compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Because `imgui-winit-support` builds `winit` with `default-features = false`, and relies on other crates to select the _mandatory_ backend(s) for a successful single compilation of `winit` in the crate graph.

PR #95, opened 2022-01-15
Add #[derive(Hash)] to the MemoryLocation enum

hrydgard:
Needed it in the buffer manager in my experimental render graph, so thought I'd just go ahead and contribute it.

Also fix some rust-2018 warnings.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -164,7 +164,7 @@ pub mod vulkan;
 #[cfg(all(windows, feature = "d3d12"))]
 pub mod d3d12;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum MemoryLocation {
     /// The allocated resource is stored at an unknown memory location; let the driver decide what's the best location
     Unknown,


Jasper-Bekkers (approved):

max-traverse:
Great!

There is a clippy error that is not relevant to your PR. Probably just a clippy update, let me check if it also fails on main.

hrydgard:
Should I try to fix the clippy, or will you take care of it?

10maurycy10:
```#[allow(non-send-fields-in-send-ty)]```

MarijnS95:
@hrydgard I don't see of these clippy lints on main besides https://github.com/Traverse-Research/gpu-allocator/pull/99, perhaps you have a stricter linter setup in `.cargo/config.toml`?

I have an old branch importing the Embark clippy lints somewhere, should perhaps dig it out and add it to this project :)

hrydgard:
Yeah, I do, but my concern is not my local clippy but your CI failing, which should be entirely unrelated :)

MarijnS95:
@hrydgard This is a lint that got default-enabled in some Rust version when you happened to open this PR, but then shortly after disabled (and it should be re-enabled now on 1.58.1?).

I can't restart GH actions but it should be gone.

I was however pointing at the elided lifetimes. We don't have `rust_2018_idioms` enabled _yet_ so our CI shouldn't fail on it. There are many more violations of it which I'm fixing in a separate PR, so perhaps those can be dropped here to make this PR purely about the change you'd like to see?

As such, to answer an implicit question: I personally prefer separate PRs over drive-by fixes, so that every PR represents a single change in essence.

hrydgard:
OK, makes sense. If you'd merge without squashing they'd be in separate commits at least, but I'll just take them out.

MarijnS95 (approved):

MarijnS95:
@hrydgard Yeah that is my general preference as well, but all Rust projects (I have come across thus far) use squash-merge and even have all the other merge strategies disabled in settings :flushed: 

I had to adapt my own strategy to this behaviour and just open multiple PRs, people tend to not review commits individually anyway. And on some of the crates I maintain it helps clean up the back-story when repeated review requests result in lots of irrelevant history churn that I can just remove when setting the final squash-commit message :partying_face:

PR #99, opened 2022-02-19
Import Embark clippy lints v5 and fix all fallout

MarijnS95:
[On newer Rust versions the linter now accurately detects fields that are only used in (specific?) `derive` macros like `Debug`, but not anywhere else in the codebase.  As such, they should still be considered unused.
](https://github.com/EmbarkStudios/rust-ecosystem/blob/64cc11e72ff4304e6c0cd0c54dbb01c6c3f151bc/lints.toml)


examples/d3d12-buffer/src/main.rs
```diff
@@ -17,16 +18,16 @@ fn create_d3d12_device(
...
-        let mut adapter4: *mut all_dxgi::IDXGIAdapter4 = std::ptr::null_mut();
+        let mut adapter4 = std::ptr::null_mut();
         let hr = unsafe {
-            dxgi_factory
-                .as_ref()
-                .unwrap()
-                .EnumAdapters1(idx, &mut adapter4 as *mut _ as *mut *mut _)
+            dxgi_factory.as_ref().unwrap().EnumAdapters1(
+                idx,
+                <*mut *mut all_dxgi::IDXGIAdapter4>::cast(&mut adapter4),
```
MarijnS95, line 25: This won't be one of my PR's if it didn't come with the typical questions of how people prefer to read Rust. I have no preference here, but we have two options:

- As in the diff selected here, remove the (now optional) type from the `let mut` binding and rely on `<*mut *mut all_dxgi::IDXGIAdapter4>::cast(&mut adapter4)` to set the source type (the `IDXGIAdapter4`) here;
- Keep the type on the `let mut`, and use `<*mut *mut _>::cast(...)` instead: this is done in a few other places in this PR;
- Be explicit about the type in both locations.

I absolutely have no preference for either. Perhaps even leaning towards the last option to be explicit about it twice.

Jasper-Bekkers (approved):

examples/d3d12-visualization/src/imgui_renderer.rs
```diff
@@ -662,7 +662,7 @@ impl ImGuiRenderer {
                 buffer.as_ref().unwrap().Map(
                     0,
                     std::ptr::null(),
-                    &mut mapped_ptr as *mut _ as *mut _,
+                    <*mut *mut _>::cast(&mut mapped_ptr),
```
max-traverse, line 665: This syntax looks a bit funny.

MarijnS95: Yeah, perhaps it needs some explanation. We're expecting `mapped_ptr` to be a `*mut T` here, so that we can get a coercion from `&mut *mut T` into `*mut *mut T`. `cast` then allows changing the type of the pointer without changing mutability (irrelevant for `*mut` since it coerces to `*const` anyway, but definitely useful in places where we'd be dealing with `*const`), probably to a `*mut c_void` in this example.

max-traverse (approved):

examples/d3d12-buffer/src/main.rs
max-traverse: I am also leaning more towards to being explicit about it twice.
I find the `<some_type>::cast(some_var)` syntax quite annoying to read.

MarijnS95: Same feeling, it's done now :)

PR #102, opened 2022-03-11
🏳️‍⚧️ Name change

manon-traverse:
Add `.mailmap` and update comments.

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -149,7 +149,7 @@ impl MemoryBlock {
 
             let allocation_flags = vk::MemoryAllocateFlags::DEVICE_ADDRESS;
             let mut flags_info = vk::MemoryAllocateFlagsInfo::builder().flags(allocation_flags);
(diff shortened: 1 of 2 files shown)


src/vulkan/mod.rs
```diff
@@ -127,9 +127,10 @@ impl Default for Allocation {
 #[derive(Debug)]
 pub(crate) struct MemoryBlock {
     pub(crate) device_memory: vk::DeviceMemory,
-    pub(crate) size: u64,
     pub(crate) mapped_ptr: *mut std::ffi::c_void,
     pub(crate) sub_allocator: Box<dyn allocator::SubAllocator>,
+    #[cfg(feature = "visualizer")]
+    pub(crate) size: u64,
```
Jasper-Bekkers, line 133: @manon-traverse looks like there are some other changes making it in too?

manon-traverse: Yeah I was branched off an old version of main 😅 
Fixing it right now

PR #105, opened 2022-03-23
Import Embark clippy lints v6 and our own internal (rustc+clippy) extensions to it

MarijnS95:


src/d3d12/mod.rs
```diff
@@ -134,7 +134,7 @@ impl<'a> AllocationCreateDesc<'a> {
 
     #[cfg(not(feature = "public-winapi"))]
     pub fn from_d3d12_resource_desc(
-        device: Dx12DevicePtr,
+        device: &Dx12DevicePtr,
```
MarijnS95, line 137: @manon-traverse what's your thought about this change? Clippy triggers here with:

```rust
warning: this argument is passed by value, but not consumed in the function body
   --> src/d3d12/mod.rs:117:17
    |
117 |         device: Dx12DevicePtr,
    |                 ^^^^^^^^^^^^^ help: consider taking a reference instead: `&Dx12DevicePtr`
    |
    = note: requested on the command line with `-W clippy::needless-pass-by-value`
help: consider marking this type as `Copy`
   --> src/d3d12/mod.rs:19:1
    |
19  | pub struct Dx12DevicePtr(pub *mut std::ffi::c_void);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
```

It is necessary because this type can't be cloned nor copied, so users need to repeatedly call `.device()` to get a new instance of this type every time they need to strangely move it into `from_d3d12_resource_desc`.

But I really don't want to mark this pointer type `Copy`, unless we add a lifetime generic (with `PhantomData`) on the `Allocator` (the one that hands out this type through `fn device()`) which is also responsible to keep the device alive (calls `.AddRef()` in `fn new()`).

It might in any case be a good idea to do that, since there is now nothing accounting for `Dx12DevicePtr` instances holding on to a pointer that might disappear when the `Allocator` is dropped.

manon-traverse (approved):

PR #107, opened 2022-03-24
Migrate core `gpu-allocator` crate and basic examples to `windows-rs`

### LINKED ISSUE #62: Switch over to windows-rs
Right now we're still reliant on the legacy `winapi` crate that's slowly being replaced with `windows-rs`, we'll need to do the same for our DirectX support.

MarijnS95:
Depends on #106
Fixes #62

As has long been planned and discussed we'd like to get rid of the ancient and barely maintained `winapi` crate, in favour of something more modern: `windows-rs` is actively developed and maintained by Microsoft themselves, and generated out of header metadata rather than written by hand.  Besides, it provides a more sophisticated and safer API for dealing with objects (abstracting away return types behind `Result`, hiding COM lifetime management, etc).

The choice for the "heavier" `windows` crate over `windows-sys` is fairly obvious: DirectX APIs are built on the COM object model which is completely unsupported in `windows-sys` making us unable to perform any method calls.  Besides, the `windows-sys` crate does very little to make the API more convenient nor safer to use, effectively taking us a step back from `winapi` instead of a step forward.

### DIFF
--- a/examples/d3d12-buffer.rs
+++ b/examples/d3d12-buffer.rs
@@ -1,3 +1,4 @@
+//! Example showcasing [`winapi`] interop with [`gpu-allocator`] which is driven by the [`windows`] crate.
 use winapi::shared::{dxgiformat, winerror};
 use winapi::um::{d3d12, d3dcommon};
 use winapi::Interface;
@@ -9,8 +10,7 @@ mod all_dxgi {
 use log::*;
 
 use gpu_allocator::d3d12::{
-    AbstractWinapiPtr, AllocationCreateDesc, Allocator, AllocatorCreateDesc, Dx12DevicePtr,
-    ResourceCategory,
+    AllocationCreateDesc, Allocator, AllocatorCreateDesc, ResourceCategory, ToWinapi, ToWindows,
 };
 use gpu_allocator::MemoryLocation;
 
@@ -68,7 +68,7 @@ fn create_d3d12_device(
                     };
                     match hr {
                         winapi::shared::winerror::S_OK => {
-                            info!("Using D3D12 feature level: {}.", feature_level_name);
+                            println!("Using D3D12 feature level: {}.", feature_level_name);
                             Some(device)
                         }
                         winapi::shared::winerror::E_NOINTERFACE => {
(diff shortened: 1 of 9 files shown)


examples/d3d12-buffer.rs
```diff
@@ -68,7 +68,7 @@ fn create_d3d12_device(
                     };
                     match hr {
                         winapi::shared::winerror::S_OK => {
-                            info!("Using D3D12 feature level: {}.", feature_level_name);
+                            println!("Using D3D12 feature level: {}.", feature_level_name);
```
MarijnS95, line 71: Changing these to `println` just like the other example, since we're not setting up any logger (`log` is just the facade) and these messages would not get printed at all.

examples/d3d12-visualization/main.rs
```diff
@@ -352,8 +352,8 @@ fn main() {
             command_list.as_mut().unwrap()
         };
 
-        let mut allocator = Allocator::new(&AllocatorCreateDesc {
-            device: Dx12DevicePtr(<*mut ID3D12Device>::cast(device)),
+        let mut allocator = Allocator::new(AllocatorCreateDesc {
+            device: (device as *const ID3D12Device).as_windows().clone(),
```
MarijnS95, line 356: I'm thinking about implementing the `ToWindows` trait for more types (`&d3d12_winapi::ID3D12Device` as borrow) and itself, and making this function accept an `impl ToWindows<ID3D12Device>` so that the user can stuff anything into here, and we'll take it from there.

However, that may be tricky for the `CreateDesc` struct setup here.

Jasper-Bekkers:
@kennykerr, Ideally I would like us to use windows-sys instead of windows-rs for this since we're mostly a downstream wrapper utility, however it looks like COM vtables aren't being built, is that an oversight or on purpose?

manon-traverse (approved):

src/d3d12/mod.rs
```diff
@@ -206,63 +255,62 @@ impl Allocation {
...
+//             chunk_id: None,
+//             offset: 0,
+//             size: 0,
+//             memory_block_index: !0,
+//             memory_type_index: !0,
+//             heap: None,
+//             name: None,
+//         }
+//     }
+// }
```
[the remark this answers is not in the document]
MarijnS95: We don't use it, and it seems wrong for end users to call this :)

PR #128, opened 2022-09-02
Initial memory visualizer for Vulkan

Jasper-Bekkers:
This actually does two things;

- It has an optimization for the callstack resolution which now only happens when it needs to instead of eagerly doing it all the time. This saves significant time in dbghelp.dll and saves some memory too since we only need to store the stack pointers instead of the string
- It includes a new allocation overview window that is just a flat list of all allocations that are currently being held by the allocator, including their name. This makes it super easy to track down where memory is being wasted.

### DIFF
--- a/src/d3d12/visualizer.rs
+++ b/src/d3d12/visualizer.rs
@@ -30,6 +30,7 @@ pub struct AllocatorVisualizer {
     selected_blocks: Vec<AllocatorVisualizerBlockWindow>,
     focus: Option<usize>,
     color_scheme: ColorScheme,
+    allocation_breakdown_sorting: Option<(Option<imgui::TableSortDirection>, usize)>,
 }
 
 fn format_heap_type(heap_type: D3D12_HEAP_TYPE) -> &'static str {
@@ -71,6 +72,7 @@ impl AllocatorVisualizer {
             selected_blocks: Vec::default(),
             focus: None,
             color_scheme: ColorScheme::default(),
+            allocation_breakdown_sorting: None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/d3d12/visualizer.rs
```diff
@@ -287,8 +289,87 @@ impl AllocatorVisualizer {
     /// - If [`true`], the widget will be drawn and an (X) closing button will be added to the widget bar.
     pub fn render(&mut self, allocator: &Allocator, ui: &imgui::Ui<'_>, opened: Option<&mut bool>) {
         if opened != Some(&mut false) {
-            self.render_main_window(ui, opened, allocator);
-            self.render_memory_block_windows(ui, allocator);
+            // self.render_main_window(ui, opened, allocator);
+            // self.render_memory_block_windows(ui, allocator);
```
manon-traverse, line 293: Do we still need this code?

src/d3d12/visualizer.rs
```diff
@@ -287,8 +289,87 @@ impl AllocatorVisualizer {
...
+
+            for memory_type in &allocator.memory_types {
+                for block in &memory_type.memory_blocks {
+                    if let Some(block) = block {
+                        allocation_report.extend_from_slice(&block.sub_allocator.report_allocations())
+                    }
+                }
+            }
+
+            if let Some(_t) = ui.begin_table_header_with_flags(
```
manon-traverse, line 315: Perhaps `.is_some()` is better here since you don't use the value in the option.

src/vulkan/visualizer.rs
```diff
@@ -275,4 +280,121 @@ impl AllocatorVisualizer {
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            total_size_in_bytes = allocation_report.iter().map(|report| report.size).sum();
+        }
+
+        let suffix = ["B", "KB", "MB", "GB", "TB"];
```
manon-traverse, line 305: Technically this should be KiB, MiB, GiB, etc
```suggestion
        let suffix = ["B", "KiB", "MiB", "GiB", "TiB"];
```

Jasper-Bekkers:
@MarijnS95 I restored the branch because this was causing internal build failures; which is why this repo is set not to delete branches when merging in the first place :-)

MarijnS95:
IMHO we should reference the `main` branch more eagerly after merging, or push a tag for as long as this rev needs to stick around, instead of leaving unused branches cluttering up which we don't have as a workflow anywhere else (and this kind of thing happened before, rev references to branches aren't reliable enough).

(Besides, it's quite tricky to get this right in the GitHub app which still jitters the "Load more items" and "Delete merged branch" button around a lot)

Now that the internal revision is updated, can the branch be re-removed? :pleading_face:

PR #130, opened 2022-09-13
🖨️ Allocator breakdown fmt

EmilioLaiso:
Add functionality to print the breakdown to the terminal instead of ImGui.

### DIFF
--- a/src/allocator/mod.rs
+++ b/src/allocator/mod.rs
@@ -86,3 +86,19 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug {
         self.allocated() == 0
     }
 }
+
+pub(crate) fn fmt_bytes(mut amount: u64) -> String {
+    let suffix = ["B", "KB", "MB", "GB", "TB"];
+
+    let mut idx = 0;
+    let mut print_amount = amount as f64;
+    loop {
+        if amount < 1024 {
+            return format!("{:.2} {}", print_amount, suffix[idx]);
+        }
+
+        print_amount = amount as f64 / 1024.0;
+        amount /= 1024;
+        idx += 1;
+    }
+}
--- a/src/allocator/mod.rs
+++ b/src/allocator/mod.rs
@@ -88,13 +88,13 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug {
 }
 
 pub(crate) fn fmt_bytes(mut amount: u64) -> String {
-    let suffix = ["B", "KB", "MB", "GB", "TB"];
+    const SUFFIX: [&str; 5] = ["B", "KB", "MB", "GB", "TB"];
 
     let mut idx = 0;
     let mut print_amount = amount as f64;
     loop {
         if amount < 1024 {
-            return format!("{:.2} {}", print_amount, suffix[idx]);
+            return format!("{:.2} {}", print_amount, SUFFIX[idx]);
         }
 
         print_amount = amount as f64 / 1024.0;
--- a/src/allocator/mod.rs
+++ b/src/allocator/mod.rs
@@ -87,6 +87,8 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug {
     }
 }
 
+pub(crate) const VISUALIZER_TABLE_MAX_ENTRY_NAME_LEN: usize = 40;
+
 pub(crate) fn fmt_bytes(mut amount: u64) -> String {
     const SUFFIX: [&str; 5] = ["B", "KB", "MB", "GB", "TB"];
 
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -749,7 +749,7 @@ impl fmt::Debug for Allocator {
 
         for memory_type in &self.memory_types {
             for block in memory_type.memory_blocks.iter().flatten() {
-                allocation_report.extend_from_slice(&block.sub_allocator.report_allocations())
+                allocation_report.extend(block.sub_allocator.report_allocations())
             }
         }
 
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -760,7 +760,10 @@ impl fmt::Debug for Allocator {
 
         const MAX_NUM_CHARACTERS: usize = 40;
 
-        f.write_str("================================================================\n")?;
+        writeln!(
+            f,
+            "================================================================"
+        )?;
         writeln!(
             f,
             "ALLOCATION BREAKDOWN ({})",
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -540,7 +540,6 @@ impl MemoryType {
     }
 }
 
-#[derive(Debug)]
 pub struct Allocator {
     device: ID3D12Device,
     debug_settings: AllocatorDebugSettings,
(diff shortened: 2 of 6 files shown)


src/d3d12/visualizer.rs
```diff
@@ -389,4 +389,43 @@ impl AllocatorVisualizer {
...
+            let num_spaces = MAX_NUM_CHARACTERS - cloned_name.len();
+            let aligning_spaces = " ".repeat(num_spaces);
+
+            println!(
+                "\t\t{}\t- {}{}\t- {}",
+                idx,
+                cloned_name,
+                aligning_spaces,
+                crate::visualizer::fmt_bytes(alloc.size)
+            );
```
manon-traverse, line 428: Should we use `info!` instead?

manon-traverse (approved):

src/d3d12/visualizer.rs
EmilioLaiso: we can, but `info!` adds a lot of unnecessary text at the beginning of each line which is really not needed when trying to print a table.

MarijnS95: I'd serialize this using a typical `fmt` API, then the caller can decide how to utilize the result instead of forcing it either to `println` or `log`.

src/d3d12/visualizer.rs
```diff
@@ -389,4 +389,43 @@ impl AllocatorVisualizer {
                 }
             });
     }
+
+    #[allow(clippy::print_stdout)]
+    pub fn debug_pring_breakdown(allocator: &Allocator) {
```
MarijnS95, line 394: Do we need the `"visualizer"` feature (and all the `imgui` deps etc that it pulls in) to support this function (i.e. stats that are only enabled on `feature = "visualizer"`)? Or can we create it as a member function on `Allocator`?

src/d3d12/mod.rs
```diff
@@ -740,6 +743,51 @@ impl Allocator {
...
+
+        let total_size_in_bytes = allocation_report.iter().map(|report| report.size).sum();
+
+        let mut allocation_report = allocation_report.iter().enumerate().collect::<Vec<_>>();
+        allocation_report.sort_by_key(|(_, alloc)| std::cmp::Reverse(alloc.size));
+
+        const MAX_NUM_CHARACTERS: usize = 40;
+
+        f.write_str("================================================================")?;
+        write!(
```
MarijnS95, line 764:
```suggestion
        writeln!(f, "================================================================")?;
        writeln!(
```

src/d3d12/mod.rs
```diff
@@ -740,6 +743,54 @@ impl Allocator {
...
+            let aligning_spaces = " ".repeat(num_spaces);
+
+            writeln!(
+                f,
+                "\t\t{}\t- {}{}\t- {}",
+                idx,
+                cloned_name,
+                aligning_spaces,
+                fmt_bytes(alloc.size)
+            )?;
```
MarijnS95, line 787:
```suggestion
            write!(
                f,
                "\t\t{}\t- {:chars$.chars$}\t- {}",
                idx,
                alloc.name,
                fmt_bytes(alloc.size),
                chars = MAX_NUM_CHARACTERS,
            )?;
```

src/vulkan/mod.rs
```diff
@@ -453,12 +455,49 @@ pub struct Allocator {
...
+            let aligning_spaces = " ".repeat(num_spaces);
+
+            writeln!(
+                f,
+                "\t\t{}\t- {}{}\t- {}",
+                idx,
+                cloned_name,
+                aligning_spaces,
+                fmt_bytes(alloc.size)
+            )?;
```
MarijnS95, line 497:
```suggestion
            write!(
                f,
                "\t\t{}\t- {:chars$.chars$}\t- {}",
                idx,
                alloc.name,
                fmt_bytes(alloc.size),
                chars = MAX_NUM_CHARACTERS,
            )?;
```

src/allocator/mod.rs
```diff
@@ -86,3 +86,19 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug {
...
 }
+
+pub(crate) fn fmt_bytes(mut amount: u64) -> String {
+    let suffix = ["B", "KB", "MB", "GB", "TB"];
+
+    let mut idx = 0;
+    let mut print_amount = amount as f64;
+    loop {
+        if amount < 1024 {
+            return format!("{:.2} {}", print_amount, suffix[idx]);
```
MarijnS95, line 97: Space needed?

src/allocator/mod.rs
```diff
@@ -86,3 +86,19 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug {
         self.allocated() == 0
     }
 }
+
+pub(crate) fn fmt_bytes(mut amount: u64) -> String {
+    let suffix = ["B", "KB", "MB", "GB", "TB"];
```
MarijnS95, line 91: const?

src/d3d12/mod.rs
```diff
@@ -740,6 +743,48 @@ impl Allocator {
...
+                allocation_report.extend_from_slice(&block.sub_allocator.report_allocations())
+            }
+        }
+
+        let total_size_in_bytes = allocation_report.iter().map(|report| report.size).sum();
+
+        let mut allocation_report = allocation_report.iter().enumerate().collect::<Vec<_>>();
+        allocation_report.sort_by_key(|(_, alloc)| std::cmp::Reverse(alloc.size));
+
+        const MAX_NUM_CHARACTERS: usize = 40;
```
MarijnS95, line 761: Generalize betweejn both backends?

src/d3d12/mod.rs
```diff
@@ -740,6 +743,48 @@ impl Allocator {
     }
 }
 
+impl fmt::Debug for Allocator {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut allocation_report = vec![];
+
+        for memory_type in &self.memory_types {
+            for block in memory_type.memory_blocks.iter().flatten() {
+                allocation_report.extend_from_slice(&block.sub_allocator.report_allocations())
```
MarijnS95, line 752: Or move the vector without borrowing?

```suggestion
                allocation_report.extend(block.sub_allocator.report_allocations())
```

src/vulkan/mod.rs
```diff
@@ -453,12 +455,43 @@ pub struct Allocator {
...
+
+        for memory_type in &self.memory_types {
+            for block in memory_type.memory_blocks.iter().flatten() {
+                allocation_report.extend_from_slice(&block.sub_allocator.report_allocations())
+            }
+        }
+
+        let total_size_in_bytes = allocation_report.iter().map(|report| report.size).sum();
+
+        let mut allocation_report = allocation_report.iter().enumerate().collect::<Vec<_>>();
```
MarijnS95, line 468: Does the `idx` here mean anything to the reader, if we sort the list anyway and the allocation is tied to a block and the allocation within instead? (both indices cannot be reconstructed from this `.enumerate()`).

src/allocator/mod.rs
EmilioLaiso: I like the space :)

MarijnS95 (approved):

PR #134, opened 2022-09-29
Add memory properties and make it accessible for Vulkan allocation

SiebenCorgie:
This pull request exposes the `vk::MemoryPropertyFlags` of an `Allocation` via the `memory_properties` function.

The reason for the request is, that I was not able to obtain the flags on the current release. If it is possible and i just did not find it, please close the request.

I need the flags when checking what the properties of an allocation using the `MemoryLocation::CpuToGpu` requirement are. Specifically, it is interesting if it contains the `DEVICE_LOCAL` bit or not. Depending on [this](https://github.com/Traverse-Research/gpu-allocator/blob/main/src/vulkan/mod.rs#L626) code the outcome can be different.

If there is anything you'd like me to add, please let me know!

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -204,7 +204,7 @@ impl Allocation {
         self.chunk_id
     }
 
-    ///Returns the [`vk::MemoryPropertyFlags`](vk::MemoryPropertyFlags) of this allocation.
+    ///Returns the [`vk::MemoryPropertyFlags`] of this allocation.
     pub fn memory_properties(&self) -> vk::MemoryPropertyFlags {
         self.memory_properties
     }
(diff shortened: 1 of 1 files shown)


manon-traverse (approved):
This looks pretty reasonable! Sorry for the long time to respond 🙈

src/vulkan/mod.rs
```diff
@@ -61,6 +62,11 @@ impl Allocation {
         self.chunk_id
     }
 
+    ///Returns the [`vk::MemoryPropertyFlags`](vk::MemoryPropertyFlags) of this allocation.
```
MarijnS95, line 65: No need for a link when the link label/text is the same (intradoc links disregard backticks):

```suggestion
    ///Returns the [`vk::MemoryPropertyFlags`] of this allocation.
```

MarijnS95 (approved):
I think this'll be interesting for more folks when there's no `HOST_CACHED` bit, and later when we relax the requirement for `HOST_COHERENT` (i.e. flush/invalidate mapped ranges manually).

SiebenCorgie:
Yeah, I came across this while trying to use/implement resizable BAR. Checking for all those flags is needed to identify if you need to flush and whether the memory is DeviceLocal.

MarijnS95 (approved):

PR #137, opened 2022-10-17
Remove `host visible but not host coherent` "warning"

EmilioLaiso:
Having this log a warning is problematic in some of our other repos. Demoting to an info :)

Edit: After some internal talks, we concluded we will get rid of this logging altogether.


MarijnS95 (approved):

MarijnS95:
It's really spamming while at it, and not something we've intended to look into for a long enough time anyway :)

MarijnS95 (approved):

PR #138, opened 2022-10-17
Add support for `presser`

fu5ha:
For CPU-to-GPU staging & update buffers, one must copy data into them. This currently requires unsafe raw pointer copies or using an unsound byte-slice interface. `presser` solves this by providing helper functions for safely copying data into raw allocated buffers while validating layout requirements and not creating invalid references to uninitialised memory.

# Implementation

This PR adds support for [`presser`](https://crates.io/crates/presser) in two ways, and for now only for Vulkan.

1. `Allocation` implements `presser::Slab` directly, meaning you can directly pass it into presser's helper functions. However, because not all `Allocation`s are valid `Slab`s, if you attempt to use a non-mapped or too-large allocation as a `Slab`, you'll cause a panic.
2. For the case where you want to handle not being able to convert to a `Slab` in some way other than panicking, `Allocation::try_as_mapped_slab` will optionally return a `MappedAllocationSlab` which pre-checks its conditions and so infallibly implements `Slab` upon succeeding.

The advantage of implementing `Slab` on `Allocation` directly, in addition to the added level of convenience it provides, is that once EmbarkStudios/presser#5 lands and we upgrade to that version here, you'll be able to share an `&Allocator` and use the `read_X` helpers from `presser` from multiple places, rather than requiring `&mut Allocator` everywhere.

# Future path to adoption

1. Land this PR
1. Design an equivalent API for the d3d12 `Allocation` type. I think this likely will look like adding a `mapped_ptr: Option<NonNull<c_void>>` field much like the `vulkan::Allocation` has, and then a `pub fn map(&mut self, resource: *mut ID3D12Resource)` which would fill in that field. But, that's for another PR's discussion...
1. At this point we could probably do a new `gpu-allocator` release just with the copy apis.
1. Land EmbarkStudios/presser#5 and upgrade `gpu-allocator` to a version `0.4` which includes those changes
1. Do another `gpu-allocator` (minor) release which adds `#[deprecated]` attributes to `mapped_slice[_mut]`, documentation detailing why they've been deprecated, and linking to an upgrade guide in some release notes or something (or just a document in the repo).
1. Once we've established that `presser` is actually replacing all use cases, do a `gpu-allocator` major release which actually removes the `#[deprecated]` methods and links to the migration guide again in the changelog/release notes.


Cargo.toml
```diff
@@ -51,6 +52,7 @@ ash-window = "0.10.0"
 raw-window-handle = "0.4"
 winit = "0.26"
 imgui-winit-support = { version = "0.8", default-features = false, features = ["winit-26"] }
+presser = "0.3"
```
MarijnS95, line 55: Add it to `required-features` for `vulkan-visualization`? It now fails to compile with:

```console
$ cargo r --example vulkan-visualization --no-default-features --features=visualizer,vulkan
error[E0599]: no method named `as_mapped_slab` found for struct `Allocation` in the current scope
...
```

I don't think you then need to have this under `[dev-dependencies]`, as it is already provided from `[dependencies]` just like some other crates.

src/vulkan/presser.rs
```diff
@@ -0,0 +1,90 @@
+use super::Allocation;
+use core::convert::TryFrom;
+
+impl Allocation {
+    /// Borrow the CPU-mapped memory that backs this allocation as a [`presser::Slab`], which you can then
+    /// use to safely copy data into the raw, potentially-uninitialized buffer.
+    ///
+    /// Returns `None` if `self.mapped_ptr()` is `None`, or if `self.size()` is > `isize::MAX` because
```
MarijnS95, line 8: Write out `larger than` or move `>` into backticks?

MarijnS95 (commented):
> This adds optional support for [`presser`](https://crates.io/crates/presser) under the `presser` feature.

Thanks! Can you describe in the description/commit, in a single line, what `presser` provides for users and why it should be used in `gpu-allocator`, without having them click through to the crate?

Depending on how this pans out we might want to land https://github.com/Traverse-Research/gpu-allocator/compare/uninit at some point or remove the currently-UB API altogether.

> Doing this is quite straightforward for Vulkan, but there didn't seem to be an equivalent for the d3d12 `Allocation` type and I'm not familiar enough with that api to say how that should be implemented

IIRC we have a `d3d12::ID3D12Resource::Map()` call in our custom implementation. I'm not sure why this isn't provided by `gpu-allocator` since this is an explicit (`vkMapMemory()`) operation on Vulkan as well... @manon-traverse any idea, perhaps worth to add in a separate PR?


> I could implement it if one told me how to acquire a mapped base pointer for some allocation...

https://github.com/Traverse-Research/gpu-allocator/blob/8844887d6458da7158e82ca5ef149ef7205e8cce/examples/d3d12-visualization/imgui_renderer.rs#L453-L456

(But in a separate, preliminary PR, if we end up going this route :wink:)

fu5ha:
> IIRC we have a `d3d12::ID3D12Resource::Map()` call in our custom implementation. I'm not sure why this isn't provided by `gpu-allocator` since this is an explicit (`vkMapMemory()`) operation on Vulkan as well... @manon-traverse any idea, perhaps worth to add in a separate PR?
> 
> >

Yeah so when looking through the d3d12 visualizer example, it seems like the main issue is that in order to get a mapped pointer you have to first create the actual `ID3D12Resource` ptr and then `Map` that, rather than mapping just the memory object that the `Allocation` knows about.. or something. So not sure exactly how an api there should look.

The `Allocation` could maybe contain an optional mapped ptr much like the vulkan `Allocation` does and then add a function like `unsafe fn map(&mut self, res: *mut ID3D12Resource)` which sets the internal mapped ptr and then allows access to the same `as_mapped_slab` api I introduced here.

Cargo.toml
fu5ha: Oh, duh, not sure why I didn't realise that was a thing.

MarijnS95:
Right, `gpu-allocator` would have to go through `CreatePlacedResource()` (and expose the returned `ID3D12Resource` for this) which seemingly isn't as clear-cut to call since it is allowed to call this multiple times in order to overlap multiple resources on the same region of heap memory.

MarijnS95:
> Not necessarily, it could require the user to do `CreatePlacedResource()` themselves and then pass the created resource back in to the `map` api on `Allocation`

Meant in the theoretical case that we extend `gpu-allocator` to perform `CreatePlacedResource()` + `Map()`, instead of accepting it as an external argument (that would then have to be asserted "somehow" to belong to the allocation), sorry for not making that more clear :)

fu5ha:
Ah right, makes sense :)

MarijnS95:
> I'm sorta unconvinced that abstracting this is much of a benefit

Agreed. I'm inclined to just make `fn mapped_slice` unsafe (since it really, _really_ is!), slap a `/// # Safety\n/// Only to be called when the memory is known to be _fully_ initialized` and leave out this whole `MaybeUninit` mess simply because _It's Complicated™_ and so easy to invalidate in the various ways you pointed out.

> I think that case of a readback buffer is probably the most compelling use case for actually tracking the initialization state of the memory as a whole since that would allow actually-safe access as a `&[u8]` for reads.

This is especially tricky since yes, you typically want to consume the data in some `&[T]` way, _and_ have to assume the GPU has written exactly that `T` to it _and_ is synchronized with the CPU timeline (and isn't mutably aliased, isn't being written to by "another thing" on the GPU timeline, ...).

Jasper-Bekkers (commented):
I'm very glad with this contribution since it's alwasy bothered me a little bit that we're effectively having to unsafe transmute our data to get it into gpu-allocator, however I have a few remarks that I'd like to share;

1. I think this is important enough to require feature parity with Dx12, it doesn't have to appear in this PR, but I will block the release of a new `gpu-allocator` version until we've landed Dx12 support as well.
2. I would like us to come up with a plan so we can start deprecating the old way of doing things and slowly move towards having `presser` be the default API here, so I'm not sure if we should hide it behind a feature toggle in the first place.

.gitignore
```diff
@@ -2,6 +2,10 @@
 # will have compiled files and executables
 /target/
 
+# A semi-common pattern is to have rust-analyzer compile into a separate target
+# directory like /tmp so that it doesn't conflict with other instances of cargo
+/tmp
+
```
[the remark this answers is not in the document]
fu5ha: Unfortunately doing it as a subdir of target/ doesn't work because the locks still conflict 😔

It is possible to make it an entirely external dir like a `~/ra_target/` or something, though I personally prefer a different one for each project 😅 but, I can also just suck it up haha

fu5ha:
Addressed Jasper's feedback and updated the original PR description with the new direction :)

fu5ha:
... github is being dumb and not letting me re-request from both of you at the same time I guess lol.

Jasper-Bekkers (approved):

src/vulkan/mod.rs
```diff
@@ -126,6 +243,55 @@ impl Default for Allocation {
     }
 }
 
+/// A wrapper struct over a borrowed [`Allocation`] that infallibly implements [`presser::Slab`].
+///
+/// This type should be acquired by calling [`Allocation::try_as_mapped_slab`].
+pub struct MappedAllocationSlab<'a> {
+    _borrowed_alloc: &'a mut Allocation,
```
djkoloski, line 250: If we don't actually use `_borrowed_alloc` then we can hold a `PhantomData<&'a mut Allocation>`. It will give us the same ownership semantics without actually using any memory.

PR #139, opened 2022-10-18
vulkan: Mark `fn mapped_(mut_)slice()` as `unsafe`

MarijnS95:
CC @fu5ha

As discussed long ago, and recently in #138, it is undefined behaviour to create or transmute to `&[u8]` when the underlying data is possibly uninit.  This also holds true for transmuting arbitrary `T: Copy` structures to `&[u8]` where eventual padding bytes are considered uninitialized, hence invalid for `u8`.

Instead of coming up with a massive safety API that distinguishes between uninitialized and initialized buffers - which turn out to be really easy to invalidate by copying structures with padding bytes - place the onus on the user to keep track of initialization status by only ever providing mapped slices in an `unsafe` context.  Users are expected to initialize the buffer using `ptr::copy(_nonoverlapping)()` when used from a CPU context instead of calling `.mapped_mut_slice()`, or switch to the new [presser] API from #138.

[presser]: https://crates.io/crates/presser

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -92,18 +92,24 @@ impl Allocation {
 
     /// Returns a valid mapped slice if the memory is host visible, otherwise it will return None.
     /// The slice already references the exact memory region of the allocation, so no offset needs to be applied.
-    pub fn mapped_slice(&self) -> Option<&[u8]> {
-        self.mapped_ptr().map(|ptr| unsafe {
-            std::slice::from_raw_parts(ptr.cast().as_ptr(), self.size as usize)
-        })
+    ///
+    /// # Safety
+    /// Only to be called when the memory is known to be _fully_ initialized.
+    pub unsafe fn mapped_slice(&self) -> Option<&[u8]> {
+        self.mapped_ptr()
+            .map(|ptr| std::slice::from_raw_parts(ptr.cast().as_ptr(), self.size as usize))
     }
 
     /// Returns a valid mapped mutable slice if the memory is host visible, otherwise it will return None.
     /// The slice already references the exact memory region of the allocation, so no offset needs to be applied.
-    pub fn mapped_slice_mut(&mut self) -> Option<&mut [u8]> {
-        self.mapped_ptr().map(|ptr| unsafe {
-            std::slice::from_raw_parts_mut(ptr.cast().as_ptr(), self.size as usize)
-        })
+    ///
+    /// # Safety
+    /// Only to be called when the memory is known to be _fully_ initialized. Use [`std::ptr::copy()`] or
+    /// [`std::ptr::copy_nonoverlapping()`] on [`mapped_ptr()`][Self::mapped_ptr()] to initialize this buffer
+    /// from the CPU instead.
+    pub unsafe fn mapped_mut_slice(&mut self) -> Option<&mut [u8]> {
+        self.mapped_ptr()
+            .map(|ptr| std::slice::from_raw_parts_mut(ptr.cast().as_ptr(), self.size as usize))
     }
 
     pub fn is_null(&self) -> bool {


src/vulkan/mod.rs
```diff
@@ -92,18 +92,24 @@ impl Allocation {
...
-    pub fn mapped_slice_mut(&mut self) -> Option<&mut [u8]> {
-        self.mapped_ptr().map(|ptr| unsafe {
-            std::slice::from_raw_parts_mut(ptr.cast().as_ptr(), self.size as usize)
-        })
+    ///
+    /// # Safety
+    /// Only to be called when the memory is known to be _fully_ initialized. Use [`std::ptr::copy()`] or
+    /// [`std::ptr::copy_nonoverlapping()`] on [`mapped_ptr()`][Self::mapped_ptr()] to initialize this buffer
+    /// from the CPU instead.
+    pub unsafe fn mapped_mut_slice(&mut self) -> Option<&mut [u8]> {
```
MarijnS95, line 110: Instead of always requiring pointers for CPU-side initialization, we can also provide a safe `fn` that returns `&(mut) [MaybeUninit<T>]`, allowing the caller to still initialize from slices more easily, without pulling in the whole (borked) "initialization tracking" from https://github.com/Traverse-Research/gpu-allocator/compare/uninit.

MarijnS95, line 110: Note that I sneaked a little `mapped_slice_mut` -> `mapped_mut_slice` in here, to better match `as_slice`/`as_mut_slice` in Rust `std`.

fu5ha:
I think that this is likely a good way forward. It might be the best to just not expose `mapped_slice` at all and only ever expose the raw base pointer, though exposing it as a `&[u8]` under an unsafe fn and making it clear the guarantees that implies is also fine. I've also thought about adding support for some unsafe functions to `presser` for casting some subsection of a `Slab` into a `&[T]` for reading. These would need to be unsafe but could help avoid some common pitfalls and list all the actual needed safety invariants in an easy-to-see place on them, but that's not a reality now so having these is probably fine.

src/vulkan/mod.rs
fu5ha: 👍 yeah, though if we wanted it to be actually safe it would still need to be `&mut [MaybeUninit<u8>]`, or check alignment and size requirements for `T` within (also doable)

MarijnS95:
It seems useful to me to have these functions available - with the right `# Safety` context - to hopefully prevent users from implementing it themselves without being aware of these problems.

That said, @fu5ha I assume for this to really be useful we should replace `u8` with `T` so that a struct with padding can still be accessed safely, as `&[T]` is safe to access whereas `&[u8]` is not.

What are your thoughts on also adding a `&mut [MaybeUninit<T>]` API (possibly both `mut` and non-`mut` variants).

fu5ha:
I'm a fan. I think the list of apis could be something like

```rust
/// `offset` is in bytes from start of allocation, `len` is the number of `T`s in the returned slice
unsafe fn mapped[_mut]slice_from_offset<T>(&[mut] self, offset: usize, len: usize) -> Option(&[mut] [T]) {}

/// Same as above but doesn't validate alignment or size requirements for you
unsafe fn mapped[_mut]_slice_from_offset_unchecked<T>(&[mut] self, offset: usize, len: usize) -> Option(&[mut] [T]) {}

/// Note this can be safe because `MaybeUninit` ships the unsafety to `assume_init` call.
/// `offset` is in bytes from start of allocation, `len` is the number of `T`s in the returned slice
fn mapped_maybe_uninit[_mut]_slice_from_offset<T>(&[mut] self, offset: usize, len: usize) -> Option(&[mut] [MaybeUninit<T>]) {}

/// Same as above but doesn't validate alignment or size requirements for you so unsafe again
unsafe fn mapped_maybe_uninit[_mut]_slice_from_offset_unchecked<T>(&[mut] self, offset: usize, len: usize) -> Option(&[mut] [MaybeUninit<T>]) {}
```

src/vulkan/mod.rs
```diff
@@ -92,18 +92,24 @@ impl Allocation {
 
     /// Returns a valid mapped slice if the memory is host visible, otherwise it will return None.
     /// The slice already references the exact memory region of the allocation, so no offset needs to be applied.
-    pub fn mapped_slice(&self) -> Option<&[u8]> {
-        self.mapped_ptr().map(|ptr| unsafe {
-            std::slice::from_raw_parts(ptr.cast().as_ptr(), self.size as usize)
-        })
+    ///
+    /// # Safety
+    /// Only to be called when the memory is known to be _fully_ initialized.
```
attackgoat, line 97: As a user of `gpu_allocator`, I can only get an `Allocation` using `Allocator::allocate`. The memory is at that point fully initialized because any bit pattern is a valid byte. Freeing likewise takes my instance, preventing unsafe usage.

Does this need to be `unsafe`?

MarijnS95: > The memory is at that point fully initialized because any bit pattern is a valid byte

According to https://crates.io/crates/presser / https://www.ralfj.de/blog/2019/07/14/uninit.html, and per the description/trigger of this PR, that is not the case.

fu5ha: > The memory is at that point fully initialized because any bit pattern is a valid byte.

But *uninit* memory is not just "any bit pattern", it is a completely separate "meta-state" about a piece of memory that means it could not just be any bit pattern but it could very well not exist at all, and thus the compiler is free to make optimizations, reorder your code, etc. based on this assumption. *reading* an uninit byte is *always* ub, even if your code is valid for any bit pattern -- see readme here https://github.com/EmbarkStudios/presser

MarijnS95: @attackgoat more concretely, will a `MaybeUninit<T>` API work for your use-case?

attackgoat: I see the danger in using data which has "random" values, especially when transmuting to `T`; but I'm not entirely convinced that this is unsafe in the context. It's more unwise. The GPU is very likely to do things to our memory as we hold it; so additional safeguards don't seem fool proof.

fu5ha: Whether the GPU is doing things or not is a wholly separate issue, though it does make things another level of unsafe. In that regard, most of the rust graphics libraries have taken the stance that the GPU overlapping access with CPU access doesn't constitute marking a function as unsafe because validating that at compile time is essentially impossible and classifying it as unsafe would mean pretty much every gpu api ever would be unsafe which just doesn't seem like it adds any value. So instead we reserve `unsafe` to demarcate when a function is *extra* unsafe and should be looked at even more carefully than just for synchronization with GPU access. Once again though, uninit =/= "random", in fact, if the GPU is writing data with uninit/padding and we are reading it back as `&[u8]` that's actually unsound whereas reading it back as `&[T]` which has the same layout as what the GPU wrote is completely sound and valid.

fu5ha:
My take is probably that things are "fine" just as is for now, people are unlikely to change to new apis between now and the near-ish future where we could replace the whole api as discussed. Especially since I don't think it'll take too long to implement those in presser and then add the APIs here on top. Plus that will be one less breaking change for users to have to update to.

src/vulkan/mod.rs
attackgoat: I prefer a strong `# Safety` warning that states the danger involved, and at least an `unsafe` marker. I don't know if I'm knowledgeable enough to know about `MaybeUnint` exactly.

fu5ha: The goal is to have an api that makes things almost-as-ergonomic while avoiding the most common safety pitfalls and calling out more explicitly the ways things can go wrong thru stronger safety warnings like you say :)

attackgoat: Yes and I'll add that I like the idea of #138 because [Screen 13](https://github.com/attackgoat/screen-13/blob/master/examples/vsm_omni.rs#L977) has the same old pattern using `bytemuck` and it would be nice to offer users a safer way of doing this.

fu5ha:
Opened https://github.com/EmbarkStudios/presser/pull/5 which implements basically the idea I discussed above, atop which it should be fairly easy to implement similar APIs in `gpu-allocator`, though probably not all of them are needed to be included directly, since we can just go through the `Slab` impl proposed in #138 for the rest.

MarijnS95:
@fu5ha Sure! This PR only took a few minutes to rig up, I'm fine with closing it and migrating the whole public mapping API to `presser` as a whole.

Fine by me to still keep the (`default` enabled) `presser` feature, and just not provide any mapping API at all if it is disabled - or keep whatever I added in this PR with the modifications (size+offset, and a safe `MaybeUninit` accessor) suggested above.

fu5ha:
Whatever you think.

I'm a bit biased but I think doing this would be a good order:

1. Keep how it is for now
2. land #138
3. land EmbarkStudios/presser#5 and release presser 0.4
4. update to presser 0.4 here and deprecate the current slice apis, recommending using the presser ones thru `as_mapped_slab` instead
5. see if it feels too awkward to go thru an extra type for all read and writes or not, if so we can add back a couple helper functions like the ones discussed above, if not we can leave as-is and fully remove the old ones

MarijnS95:
Sounds good!

CC @manon-traverse @Jasper-Bekkers :)

Jasper-Bekkers:
> We could keep the (default enabled) presser feature in https://github.com/Traverse-Research/gpu-allocator/pull/138, and just not provide any mapping API at all if it is disabled

I think we should always provide a mapping api in order to not break this crate; however, I think we should support `presser` as a first class citizen and potentially even not provide any other paths. However, if we go that route, we should make sure that it's minimal impact on exisiting users, and has great upgrade documentation.

MarijnS95:
> I think we should always provide a mapping api in order to not break this crate

Yeah, it would always go through `presser` in that case, and we should omit the feature flag altogether (if users really are against using `presser` in the public API, we could provide minimal mapping functions that are implemented on top of `presser` internally :grimacing:).

---

For this to work out I suggest we follow @fu5ha's points above - drop this PR - and use at least the examples in this repo _and_ our codebase to test-drive a full public-API migration to `presser` when point 4 is finished. We can always make beta releases or provide a non-breaking release with `#[deprecated]` statements, just to see how it all pans out without affecting anyone.

fu5ha:
Sounds good to me. We can/will also use our internal codebase to test things out!

djkoloski:
Would it be worth releasing a version that deprecates `mapped_(mut_)?slice()`? That would ensure that anyone using the current API should be using `mapped_ptr()` and `size()` instead.

fu5ha:
@djkoloski yeah.. this PR is likely to be closed rather than merge now, see the current plan described in first comment of #138

MarijnS95:
@djkoloski That is effectively - exactly - what I wrote two posts above yours :wink:. This is planned, I just have had too much on my plate to look into https://github.com/EmbarkStudios/presser/pull/5 followed by #138, apologies @fu5ha I'll queue it up asap!

MarijnS95:
I am reopening this PR, if not in the least for tracking purposes. While the presser changes were merged as-is (adding lots of documentation and a crate dependency which I have yet to digest, for what is effectively a raw pointer API), while leaving the original UB `fn mapped_(mut_)slice()` in place, and leaving it safe to call.

We'd have to remove this API or mark it `unsafe`, as well as provide the proposed `MaybeUninit` convenience accessors.

PR #148, opened 2022-12-21
Add support for dedicated allocations and committed resource tracking

janie177:
### Changes
This PR adds the option to create dedicated allocations in the Vulkan allocator and CommittedResources in the Dx12 allocator. These types of allocations are managed by the driver, which allows it to do special optimizations (🧙‍♂️ magic). 

### Vulkan
In the Vulkan allocator, you must now pass an extra bool `dedicated_allocation` into the `allocate` function. Vulkan 1.1 is required for this to be available, though there is an extension for Vulkan 1.0 available (which I didn't integrate, but I could do that if we want to support it).

The visualizer now displays whether a memory block is backed by a dedicated allocation or not.

### D3D12
In DirectX, dedicated allocations are hidden behind the `CreateCommittedResource()` call. It is not possible to access these allocations, so I've added a `create_resource` function to the dx12 allocator which will return a `Resource` object. You can decide whether you want this to be a `Placed` or `Committed` resource, and the rest is handled under the hood for you. When freeing a resource, you have to call the `free_resource` function on the allocator. 

I have added a special section that displays the total amount of committed resources and their total size per memory type to the visualizer. 

### Testing
I have tested these changes in the dx12 and vulkan backends of the Breda framework. Performance on vulkan and dx12 increased by about 10-fold in our ray-tracing shaders when marking render targets as dedicated resources.

### DIFF
--- a/examples/vulkan-buffer.rs
+++ b/examples/vulkan-buffer.rs
@@ -111,6 +111,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
+                dedicated_allocation: false,
                 name: "Test allocation (Gpu Only)",
             })
             .unwrap();
@@ -143,6 +144,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
+                dedicated_allocation: false,
                 name: "Test allocation (Cpu to Gpu)",
             })
             .unwrap();
@@ -175,6 +177,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
+                dedicated_allocation: false,
                 name: "Test allocation (Gpu to Cpu)",
             })
             .unwrap();
--- a/examples/vulkan-buffer.rs
+++ b/examples/vulkan-buffer.rs
@@ -3,7 +3,9 @@ use ash::vk;
 use std::default::Default;
 use std::ffi::CString;
 
-use gpu_allocator::vulkan::{AllocationCreateDesc, Allocator, AllocatorCreateDesc};
+use gpu_allocator::vulkan::{
+    AllocationCreateDesc, AllocationScheme, Allocator, AllocatorCreateDesc,
+};
 use gpu_allocator::MemoryLocation;
 
 fn main() {
@@ -111,7 +113,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
-                dedicated_allocation: false,
+                allocation_scheme: AllocationScheme::GpuAllocatorManaged,
                 name: "Test allocation (Gpu Only)",
             })
             .unwrap();
@@ -144,7 +146,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
-                dedicated_allocation: false,
+                allocation_scheme: AllocationScheme::GpuAllocatorManaged,
                 name: "Test allocation (Cpu to Gpu)",
             })
             .unwrap();
@@ -177,7 +179,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
-                dedicated_allocation: false,
+                allocation_scheme: AllocationScheme::GpuAllocatorManaged,
                 name: "Test allocation (Gpu to Cpu)",
             })
             .unwrap();
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -24,6 +24,16 @@ mod public_winapi {
         fn as_windows(&self) -> &T;
     }
 
+    impl ToWinapi<winapi_d3d12::ID3D12Resource> for ID3D12Resource {
+        fn as_winapi(&self) -> *const winapi_d3d12::ID3D12Resource {
+            unsafe { std::mem::transmute_copy(self) }
+        }
+
+        fn as_winapi_mut(&mut self) -> *mut winapi_d3d12::ID3D12Resource {
+            unsafe { std::mem::transmute_copy(self) }
+        }
+    }
+
     impl ToWinapi<winapi_d3d12::ID3D12Device> for ID3D12Device {
         fn as_winapi(&self) -> *const winapi_d3d12::ID3D12Device {
             unsafe { std::mem::transmute_copy(self) }
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -246,12 +246,12 @@ pub struct Allocation {
 }
 
 pub enum ResourceType<'a> {
-    /// Allocation equivalent to Dx12's CommittedResources.
+    /// Allocation equivalent to Dx12's CommittedResource.
     Committed {
         heap_properties: &'a D3D12_HEAP_PROPERTIES,
         heap_flags: D3D12_HEAP_FLAGS,
     },
-    /// Allocation equivalent to Dx12's PlacedResources.
+    /// Allocation equivalent to Dx12's PlacedResource.
     Placed,
 }
 
@@ -260,12 +260,12 @@ pub struct Resource {
     pub allocation: Option<Allocation>,
     pub resource: ManuallyDrop<ID3D12Resource>,
     pub memory_location: MemoryLocation,
-    heap_category: HeapCategory,
+    memory_type_index: Option<usize>,
     pub size: u64,
 }
 
 #[derive(Debug)]
-pub struct CommittedAllocationData {
+pub struct CommittedAllocationStatistics {
     pub num_allocations: usize,
     pub total_size: u64,
 }
@@ -368,7 +368,7 @@ unsafe impl Sync for MemoryBlock {}
 
 struct MemoryType {
     memory_blocks: Vec<Option<MemoryBlock>>,
-    committed_allocations: CommittedAllocationData,
+    committed_allocations: CommittedAllocationStatistics,
     memory_location: MemoryLocation,
     heap_category: HeapCategory,
     heap_properties: D3D12_HEAP_PROPERTIES,
@@ -687,7 +687,7 @@ impl Allocator {
                     heap_properties,
                     memory_type_index: i,
                     active_general_blocks: 0,
-                    committed_allocations: CommittedAllocationData {
+                    committed_allocations: CommittedAllocationStatistics {
                         num_allocations: 0,
                         total_size: 0,
                     },
@@ -796,7 +796,7 @@ impl Allocator {
     }
 
     /// Create a resource according to the provided parameters.
-    /// Created resources should be freed at the end of their lifetime by calling [`Self::free_resource`];
+    /// Created resources should be freed at the end of their lifetime by calling [`Self::free_resource()`].
     pub fn create_resource(&mut self, desc: &ResourceCreateDesc<'_>) -> Result<Resource> {
         match desc.resource_type {
             ResourceType::Committed {
@@ -806,7 +806,7 @@ impl Allocator {
                 let mut result: Option<ID3D12Resource> = None;
 
                 let clear_value: Option<*const D3D12_CLEAR_VALUE> =
-                    desc.clear_value.map(|v| <*const _>::cast(v));
+                    desc.clear_value.map(|v| -> *const _ { v });
 
                 if let Err(err) = unsafe {
                     self.device.CreateCommittedResource(
@@ -828,16 +828,6 @@ impl Allocator {
                             .GetResourceAllocationInfo(0, &[*desc.resource_desc])
                     };
 
-                    let resource_category = if (desc.resource_desc.Flags
-                        & (D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET
-                            | D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))
-                        != D3D12_RESOURCE_FLAG_NONE
-                    {
-                        ResourceCategory::RtvDsvTexture
-                    } else {
-                        ResourceCategory::OtherTexture
-                    };
-
                     let memory_type = self
                         .memory_types
                         .iter_mut()
@@ -848,7 +838,7 @@ impl Allocator {
 
                             let is_category_compatible = memory_type.heap_category
                                 == HeapCategory::All
-                                || memory_type.heap_category == resource_category.into();
+                                || memory_type.heap_category == desc.resource_category.into();
 
                             is_location_compatible && is_category_compatible
                         })
@@ -861,8 +851,8 @@ impl Allocator {
                         allocation: None,
                         resource: ManuallyDrop::new(resource),
                         size: allocation_info.SizeInBytes,
-                        heap_category: resource_category.into(),
                         memory_location: desc.memory_location,
+                        memory_type_index: Some(memory_type.memory_type_index),
                     })
                 }
             }
@@ -904,8 +894,8 @@ impl Allocator {
                         allocation: Some(allocation),
                         resource: ManuallyDrop::new(resource),
                         size,
-                        heap_category: desc.resource_category.into(),
                         memory_location: desc.memory_location,
+                        memory_type_index: None,
                     })
                 }
             }
@@ -914,30 +904,18 @@ impl Allocator {
 
     /// Free a resource and its memory.
     pub fn free_resource(&mut self, resource: Resource) -> Result<()> {
-        let memory_location = resource.memory_location;
-        let heap_category = resource.heap_category;
         let _ = ManuallyDrop::into_inner(resource.resource);
         if let Some(allocation) = resource.allocation {
             self.free(allocation)
         } else {
-            // Committed resource
-            let memory_type = self
-                .memory_types
-                .iter_mut()
-                .find(|memory_type| {
-                    let is_location_compatible = memory_location == MemoryLocation::Unknown
-                        || memory_location == memory_type.memory_location;
+            // Committed resources are implicitly dropped when their refcount reaches 0.
+            // We only have to change the allocation and size tracking.
+            if let Some(memory_type_index) = resource.memory_type_index {
+                let memory_type = &mut self.memory_types[memory_type_index];
 
-                    let is_category_compatible = memory_type.heap_category == HeapCategory::All
-                        || memory_type.heap_category == heap_category;
-
-                    is_location_compatible && is_category_compatible
-                })
-                .ok_or(AllocationError::NoCompatibleMemoryTypeFound)?;
-
-            memory_type.committed_allocations.num_allocations -= 1;
-            memory_type.committed_allocations.total_size -= resource.size;
-
+                memory_type.committed_allocations.num_allocations -= 1;
+                memory_type.committed_allocations.total_size -= resource.size;
+            }
             Ok(())
         }
     }
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -97,6 +97,17 @@ pub enum ResourceCategory {
     OtherTexture,
 }
 
+#[derive(Clone, Copy)]
+pub struct ResourceCreateDesc<'a> {
+    pub name: &'a str,
+    pub memory_location: MemoryLocation,
+    pub resource_category: ResourceCategory,
+    pub resource_desc: &'a D3D12_RESOURCE_DESC,
+    pub clear_value: Option<&'a D3D12_CLEAR_VALUE>,
+    pub initial_state: D3D12_RESOURCE_STATES,
+    pub resource_type: &'a ResourceType<'a>,
+}
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum HeapCategory {
     All,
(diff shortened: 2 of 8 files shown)


Jasper-Bekkers (approved):

src/d3d12/mod.rs
```diff
@@ -224,6 +245,31 @@ pub struct Allocation {
     name: Option<Box<str>>,
 }
 
+pub enum ResourceType<'a> {
+    /// Allocation equivalent to Dx12's CommittedResources.
```
MarijnS95, line 249: Drop `-s` from `CommittedResources`? Same for `PlacedResources` below?

src/d3d12/mod.rs
```diff
@@ -743,6 +794,153 @@ impl Allocator {
             }
         }
     }
+
+    /// Create a resource according to the provided parameters.
+    /// Created resources should be freed at the end of their lifetime by calling [`Self::free_resource`];
```
MarijnS95, line 799:
```suggestion
    /// Created resources should be freed at the end of their lifetime by calling [`Self::free_resource()`].
```

src/vulkan/mod.rs
```diff
@@ -25,6 +25,9 @@ pub struct AllocationCreateDesc<'a> {
     pub location: MemoryLocation,
     /// If the resource is linear (buffer / linear texture) or a regular (tiled) texture.
     pub linear: bool,
+    /// True if the memory should be allocated in dedicated system memory.
+    /// If false, the memory will be placed in a potentially shared allocation block.
+    pub dedicated_allocation: bool,
```
MarijnS95, line 30: No need to write out "true"/"false" just like the doc comment above, that's inherent to the type (and we'll drop `if` as well). How about:

```suggestion
    /// This allocation should be placed in dedicated system memory according to the driver.
    /// It will otherwise be placed in a potentially shared allocation block.
    pub dedicated_allocation: bool,
```

src/vulkan/mod.rs
```diff
@@ -122,16 +131,28 @@ impl Default for Allocation {
...
+            dedicated_allocation: false,
         }
     }
 }
 
+pub enum BlockType {
+    /// This block only contains a single resource.
+    Dedicated {
+        /// When true, the allocation behind this block will be a "dedicated allocation".
+        dedicated_allocation: bool,
```
MarijnS95, line 143: And what if this is `false`, but the `BlockType` is still `Dedicated`?

src/vulkan/visualizer.rs
```diff
@@ -218,8 +221,7 @@ impl AllocatorVisualizer {
                     // Imgui can actually modify this number to be out of bounds, so we will clamp manually.
                     window.bytes_per_unit = window
                         .bytes_per_unit
-                        .min(BYTES_PER_UNIT_MAX)
-                        .max(BYTES_PER_UNIT_MIN);
+                        .clamp(BYTES_PER_UNIT_MAX, BYTES_PER_UNIT_MIN);
```
MarijnS95, line 224: Lol
```suggestion
                        .clamp(BYTES_PER_UNIT_MIN, BYTES_PER_UNIT_MAX);
```

src/vulkan/mod.rs
```diff
@@ -251,14 +291,22 @@ impl MemoryType {
         let size = desc.requirements.size;
         let alignment = desc.requirements.alignment;
 
-        // Create a dedicated block for large memory allocations
-        if size > memblock_size {
+        let block_type = if size > memblock_size || desc.dedicated_allocation {
+            BlockType::Dedicated {
+                dedicated_allocation: desc.dedicated_allocation,
```
MarijnS95, line 296: I think this bit of code clarifies my question above:

- `Dedicated` -> use our dedicated block allocator;
- `dedicated_allocation`: `true` if requested by the user, `false` if not requested but determined by `gpu-allocator` because the size request is larger than a memory block. Also does _not_ result in using Vulkan's `vk::MemoryDedicatedAllocateInfo`.

Can our API represent and document this more clearly, by improving the name and documentation of `dedicated_allocation` or perhaps using a separate `BlockType` variant altogether?

src/vulkan/mod.rs
```diff
@@ -303,6 +351,7 @@ impl MemoryType {
                 device_memory: mem_block.device_memory,
                 mapped_ptr: std::ptr::NonNull::new(mem_block.mapped_ptr),
                 name: Some(desc.name.into()),
+                dedicated_allocation: false,
```
MarijnS95, line 354: Are you sure this shouldn't be `block_type`'s `dedicated_allocation` member? Otherwise add a comment explaining why not.

src/vulkan/mod.rs
```diff
@@ -334,6 +383,7 @@ impl MemoryType {
                             memory_type_index: self.memory_type_index,
                             device_memory: mem_block.device_memory,
                             mapped_ptr,
+                            dedicated_allocation: desc.dedicated_allocation,
```
MarijnS95, line 386: And this could be false, since it's unreachable? When `desc.dedicated_allocation` is true you end up setting `BlockType::Dedicated {}` and end up in the `if` + `return` right above?

src/d3d12/visualizer.rs
```diff
@@ -242,8 +250,7 @@ impl AllocatorVisualizer {
                     // Imgui can actually modify this number to be out of bounds, so we will clamp manually.
                     window.bytes_per_unit = window
                         .bytes_per_unit
-                        .min(BYTES_PER_UNIT_MAX)
-                        .max(BYTES_PER_UNIT_MIN);
+                        .clamp(BYTES_PER_UNIT_MAX, BYTES_PER_UNIT_MIN);
```
MarijnS95, line 253: Lol ^2?

examples/vulkan-buffer.rs
```diff
@@ -111,6 +111,7 @@ fn main() {
                 requirements,
                 location,
                 linear: true,
+                dedicated_allocation: false,
```
MarijnS95, line 114: Should we have an example/test using `true` here, in both the Vulkan and DX12 examples?

src/d3d12/mod.rs
```diff
@@ -743,6 +794,153 @@ impl Allocator {
...
+    pub fn create_resource(&mut self, desc: &ResourceCreateDesc<'_>) -> Result<Resource> {
+        match desc.resource_type {
+            ResourceType::Committed {
+                heap_properties,
+                heap_flags,
+            } => {
+                let mut result: Option<ID3D12Resource> = None;
+
+                let clear_value: Option<*const D3D12_CLEAR_VALUE> =
+                    desc.clear_value.map(|v| <*const _>::cast(v));
```
MarijnS95, line 809: It's a bit unfortunate since this doesn't require a cast, only a coercion from a borrow to pointer type. The cast now allows inadvertently changing the type.

We can get around this by setting the return-type of the closure to force the coercion, while still enforcing the types to remain identical. You can even inline this below to force the lifetime:

```rust
desc.clear_value.map(|v| -> *const _ { v })
```

src/d3d12/mod.rs
```diff
@@ -743,6 +794,153 @@ impl Allocator {
...
+                    self.device.CreateCommittedResource(
+                        *heap_properties,
+                        *heap_flags,
+                        desc.resource_desc,
+                        desc.initial_state,
+                        clear_value,
+                        &mut result,
+                    )
+                } {
+                    Err(AllocationError::Internal(err.message().to_string()))
```
MarijnS95, line 821: I don't think this is how we format Windows-rs errors elsewhere in our d3d12/dx12 backend?

src/d3d12/mod.rs
```diff
@@ -224,6 +245,31 @@ pub struct Allocation {
...
+pub struct Resource {
+    pub allocation: Option<Allocation>,
+    pub resource: ManuallyDrop<ID3D12Resource>,
+    pub memory_location: MemoryLocation,
+    heap_category: HeapCategory,
+    pub size: u64,
+}
+
+#[derive(Debug)]
+pub struct CommittedAllocationData {
```
MarijnS95, line 268: Statistics?

src/d3d12/mod.rs
```diff
@@ -743,6 +794,153 @@ impl Allocator {
...
+
+                    let is_category_compatible = memory_type.heap_category == HeapCategory::All
+                        || memory_type.heap_category == heap_category;
+
+                    is_location_compatible && is_category_compatible
+                })
+                .ok_or(AllocationError::NoCompatibleMemoryTypeFound)?;
+
+            memory_type.committed_allocations.num_allocations -= 1;
+            memory_type.committed_allocations.total_size -= resource.size;
```
MarijnS95, line 939: Or store `memory_type_index` so that you don't have to do this full compatibility lookup again?

src/d3d12/visualizer.rs
```diff
@@ -140,6 +140,14 @@ impl AllocatorVisualizer {
                             ));
                             ui.text(format!("total block size: {} KiB", total_block_size / 1024));
                             ui.text(format!("total allocated:  {} KiB", total_allocated / 1024));
+                            ui.text(format!(
+                                "num committed resource allocations: {}",
+                                mem_type.committed_allocations.num_allocations
+                            ));
+                            ui.text(format!(
+                                "total committed resource allocations size: {} KiB",
```
MarijnS95, line 148:
```suggestion
                                "total committed resource allocation size: {} KiB",
```

(or remove `size`, that's pretty clear from `KiB`?)

src/d3d12/visualizer.rs
```diff
@@ -140,6 +140,14 @@ impl AllocatorVisualizer {
                             ));
                             ui.text(format!("total block size: {} KiB", total_block_size / 1024));
                             ui.text(format!("total allocated:  {} KiB", total_allocated / 1024));
+                            ui.text(format!(
+                                "num committed resource allocations: {}",
```
MarijnS95, line 144: No need for `num` suffix?

src/vulkan/mod.rs
```diff
@@ -157,6 +185,16 @@ impl MemoryBlock {
                 alloc_info
             };
 
+            // Flag the memory as dedicated if required.
+            let mut dedicated_memory_info = vk::MemoryDedicatedAllocateInfo::builder();
+            let alloc_info = if dedicated_allocation {
+                // We don't yet know the buffer nor the image that will be placed in this memory.
+                // The user has to bind them after the allocation has been made.
+                alloc_info.push_next(&mut dedicated_memory_info)
```
MarijnS95, line 193: Maybe clarify that this comment corresponds to members of `vk::MemoryDedicatedAllocateInfo` that have to be filled in? How does the caller even fill them in, since this struct is passed directly into `device.allocate_memory()`?

src/vulkan/mod.rs
```diff
@@ -140,9 +161,16 @@ impl MemoryBlock {
...
         mem_type_index: usize,
         mapped: bool,
-        dedicated: bool,
+        block_type: &BlockType,
         buffer_device_address: bool,
     ) -> Result<Self> {
+        let dedicated_allocation = match block_type {
+            BlockType::Dedicated {
+                dedicated_allocation,
+            } => *dedicated_allocation,
```
MarijnS95, line 170: So if it's `Dedicated`, but `dedicated_allocation` is `false`, this isn't dedicated?

src/d3d12/visualizer.rs
janie177: Clippy was complaining about this earlier so I applied the suggestion it gave me, but yes that's a big lol. Fixed it in both places now.

src/vulkan/mod.rs
janie177: I chose to put these in the same block type because a dedicated allocation will always require a dedicated block. Therefore adding it as an inner variable to dedicated blocks made sense to me. 

I will document the user exposed function a bit more to clarify what it does though!

src/vulkan/mod.rs
janie177: Good catch, I changed this in the wrong place when I was refactoring some stuff clippy complained about! This would have caused the visualizer to display incorrect information.

src/vulkan/mod.rs
janie177: Yes exactly, this is where I meant to place the false that was incorrectly set in the block above. I must have switched them up during the refactor.

examples/vulkan-buffer.rs
janie177: It would be nice to have, so I may add one if I have some time. For now I at least tested the changes in Breda (before clippy refactors) so I'll do another round of testing at least before we merge this.

src/d3d12/mod.rs
janie177: Ohh that is interesting! Personally I'm still a bigger fan of being more explicit about types and just casting to the right one, but if clippy demands it then it shall be so. Added your suggestion as it's at least cleaner syntax wise.

src/d3d12/mod.rs
janie177: I thought about doing that, but wasn't sure if we could assume that the memory type indices stay persistent. But I see now that they actually do, so I'll add them to the resource!

src/vulkan/mod.rs
janie177: The docs don't mention that these have to be set, only that if one is set the other must be NULL. I didn't get any validation warnings and all seems to work as expected with dedicated allocations for render targets. I'll update the comment to make this more clear 👍

src/d3d12/visualizer.rs
MarijnS95: The suggestion it gave you would have been the right way around, at least it is for me. Maybe you just replaced `.min(` with `.clamp(` and `)\n.max(` with `, ` :joy: 

```rust
warning: clamp-like pattern without using clamp function
   --> src/vulkan/visualizer.rs:219:45
    |
219 |                       window.bytes_per_unit = window
    |  _____________________________________________^
220 | |                         .bytes_per_unit
221 | |                         .min(BYTES_PER_UNIT_MAX)
222 | |                         .max(BYTES_PER_UNIT_MIN);
    | |________________________________________________^
    |
    = note: clamp will panic if max < min
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_clamp
    = note: `-W clippy::manual-clamp` implied by `-W clippy::all`
help: replace with clamp
    |
219 ~                     window.bytes_per_unit = window
220 ~                         .bytes_per_unit.clamp(BYTES_PER_UNIT_MIN, BYTES_PER_UNIT_MAX);
    |

warning: clamp-like pattern without using clamp function
   --> src/d3d12/visualizer.rs:243:45
    |
243 |                       window.bytes_per_unit = window
    |  _____________________________________________^
244 | |                         .bytes_per_unit
245 | |                         .min(BYTES_PER_UNIT_MAX)
246 | |                         .max(BYTES_PER_UNIT_MIN);
    | |________________________________________________^
    |
    = note: clamp will panic if max < min
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_clamp
help: replace with clamp
    |
243 ~                     window.bytes_per_unit = window
244 ~                         .bytes_per_unit.clamp(BYTES_PER_UNIT_MIN, BYTES_PER_UNIT_MAX);
    |
```

src/d3d12/mod.rs
MarijnS95: Exactly that, and using a `::cast()` doesn't at all make this clear since you're not casting anything, you're only coercing `&T` to `*const T` and then casting to `*const T` (it is the same type). There's no clean way to force the coercion, and Rust doesn't seem to perform it for you when wrapped inside an `Option` via a `.map(|v| v)`... :(

src/vulkan/mod.rs
MarijnS95: That's seriously odd, there's no VUID to state that either should be non-null, but the description clearly mentions:

> then that structure includes a handle of the sole buffer or image resource that the memory can be bound to.

How can the driver know what the optimal memory for a thing would be... If you don't tell it about "thing"?

src/d3d12/mod.rs
MarijnS95: We don't call `.message()` anywhere else for these, it's always something like:

https://github.com/Traverse-Research/gpu-allocator/blob/abcc0ccb6ca71503f1fcbf472eff863474fb8157/src/d3d12/mod.rs#L340-L343

Where we explain _what_ function was called before `Display`-formatting `err`.

MarijnS95: That said I'd love to start moving these to `anyhow` and turn this into a `let x = WindowsFunction().context("While calling WindowsFunction")?;`.

src/vulkan/mod.rs
```diff
@@ -15,6 +15,19 @@ use crate::{
...
+    /// The driver will be able to perform optimizations in some cases with this type of allocation.
+    DedicatedBuffer { buffer: vk::Buffer },
+    /// This allocation will be for a image, and it will be driver managed using vulkans DedicatedAllocation feature.
+    /// The driver will be able to perform optimizations in some cases with this type of allocation.
+    DedicatedImage { image: vk::Image },
+    /// This allocation will be managed by the GPU allocator.
+    /// It is possible that the allocation will reside in an allocation block shared with other resources.
+    GpuAllocatorManaged,
+}
+
```
janie177, line 30: @MarijnS95 What do you think about this for the user exposed interface? 
We can now also correctly pass the buffer and image when required for dedicated allocations. I got rid of the `BlockType` as that was essentially serving the same purpose, just with less configurability and under the hood.

src/vulkan/mod.rs
Jasper-Bekkers: Yeah we need to disambiguate those terms.

src/vulkan/mod.rs
janie177: Yeah really strange, I changed the UI now so that the user has to provide the image or buffer that they want the memory to be for when they do a dedicated allocation.

src/vulkan/mod.rs
janie177: Removed this entirely now and replaced it with a something the user has control over 👍

src/vulkan/mod.rs
janie177: I've changed the UI so this is no longer there.

src/d3d12/mod.rs
```diff
@@ -743,6 +794,131 @@ impl Allocator {
             }
         }
     }
+
+    /// Create a resource according to the provided parameters.
+    /// Created resources should be freed at the end of their lifetime by calling [`Self::free_resource()`].
```
MarijnS95, line 799: Do we need a `Drop` destructor to call this or print a warning when it was not called?

src/d3d12/mod.rs
MarijnS95: @janie177 :eyes:

src/d3d12/mod.rs
```diff
@@ -743,6 +794,131 @@ impl Allocator {
...
+    }
+
+    /// Free a resource and its memory.
+    pub fn free_resource(&mut self, resource: Resource) -> Result<()> {
+        let _ = ManuallyDrop::into_inner(resource.resource);
+        if let Some(allocation) = resource.allocation {
+            self.free(allocation)
+        } else {
+            // Committed resources are implicitly dropped when their refcount reaches 0.
+            // We only have to change the allocation and size tracking.
```
MarijnS95, line 912: Well you're explicitly dropping `resource` with `ManuallyDrop` right above... It's just that there's no `allocation` to clean up here.

src/vulkan/mod.rs
MarijnS95: Interface looks good, though I'd have reworded some of the comments:

```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub enum AllocationScheme {
    /// Perform a dedicated, driver-managed allocation for the given buffer, allowing
    /// it to perform optimizations on this type of allocation.
    DedicatedBuffer { buffer: vk::Buffer },
    /// Perform a dedicated, driver-managed allocation for the given image, allowing
    /// it to perform optimizations on this type of allocation.
    DedicatedImage { image: vk::Image },
    /// The memory for this resource will be allocated and managed by gpu-allocator.
    /// It is possible that the allocation will reside in [`vk::Memory`] shared with other resources.
    GpuAllocatorManaged,
}

```

src/vulkan/mod.rs
```diff
@@ -140,9 +163,11 @@ impl MemoryBlock {
         size: u64,
         mem_type_index: usize,
         mapped: bool,
-        dedicated: bool,
         buffer_device_address: bool,
+        allocation_scheme: &AllocationScheme,
```
MarijnS95, line 167: This is a cheap `Copy` type, just pass-by-value. The user will likely ad-hoc create it anyway.

src/vulkan/mod.rs
```diff
@@ -15,6 +15,19 @@ use crate::{
...
 };
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum AllocationScheme {
+    /// This allocation will be for a buffer, and it will be driver managed using vulkans DedicatedAllocation feature.
+    /// The driver will be able to perform optimizations in some cases with this type of allocation.
+    DedicatedBuffer { buffer: vk::Buffer },
+    /// This allocation will be for a image, and it will be driver managed using vulkans DedicatedAllocation feature.
+    /// The driver will be able to perform optimizations in some cases with this type of allocation.
+    DedicatedImage { image: vk::Image },
```
MarijnS95, line 25: Also feel free to just use a tuple-variant here, the name of the variant conveys the same name as the only named field here.

src/vulkan/mod.rs
```diff
@@ -183,17 +222,21 @@ impl MemoryBlock {
             std::ptr::null_mut()
         };
 
-        let sub_allocator: Box<dyn allocator::SubAllocator> = if dedicated {
-            Box::new(allocator::DedicatedBlockAllocator::new(size))
-        } else {
-            Box::new(allocator::FreeListAllocator::new(size))
-        };
+        let sub_allocator: Box<dyn allocator::SubAllocator> =
+            if !matches!(allocation_scheme, AllocationScheme::GpuAllocatorManaged)
```
MarijnS95, line 226: Just use `==` :)

PR #151, opened 2023-01-16
Resource cleanup

janie177:
### Changes
This addresses some feedback on my last PR (#148). Comments have been adjusted and function signatures were slightly changed.

The biggest change is that dx12 resources are no longer manually dropped. They can now be accessed through the `.resource()` function that is called on `Resource`. If the user does not call `Allocator::free_resource()` before dropping a resource, a warning will be printed.

### DIFF
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -2,7 +2,7 @@
 
 use std::{fmt, mem::ManuallyDrop};
 
-use log::{debug, Level};
+use log::{debug, warn, Level};
 
 use windows::Win32::{Foundation::E_OUTOFMEMORY, Graphics::Direct3D12::*};
 
@@ -258,10 +258,24 @@ pub enum ResourceType<'a> {
 #[derive(Debug)]
 pub struct Resource {
     pub allocation: Option<Allocation>,
-    pub resource: ManuallyDrop<ID3D12Resource>,
+    resource: Option<ID3D12Resource>,
     pub memory_location: MemoryLocation,
     memory_type_index: Option<usize>,
     pub size: u64,
+}
+
+impl Resource {
+    pub fn resource(&self) -> &ID3D12Resource {
+        self.resource.as_ref().expect("Resource was already freed.")
+    }
+}
+
+impl Drop for Resource {
+    fn drop(&mut self) {
+        if self.resource.is_some() {
+            warn!("Dropping resource that was not freed. Was `Allocator::free_resource()` not called?")
+        }
+    }
 }
 
 #[derive(Debug)]
@@ -849,7 +863,7 @@ impl Allocator {
 
                     Ok(Resource {
                         allocation: None,
-                        resource: ManuallyDrop::new(resource),
+                        resource: Some(resource),
                         size: allocation_info.SizeInBytes,
                         memory_location: desc.memory_location,
                         memory_type_index: Some(memory_type.memory_type_index),
@@ -892,7 +906,7 @@ impl Allocator {
                     let size = allocation.size();
                     Ok(Resource {
                         allocation: Some(allocation),
-                        resource: ManuallyDrop::new(resource),
+                        resource: Some(resource),
                         size,
                         memory_location: desc.memory_location,
                         memory_type_index: None,
@@ -903,13 +917,13 @@ impl Allocator {
     }
 
     /// Free a resource and its memory.
-    pub fn free_resource(&mut self, resource: Resource) -> Result<()> {
-        let _ = ManuallyDrop::into_inner(resource.resource);
-        if let Some(allocation) = resource.allocation {
+    pub fn free_resource(&mut self, mut resource: Resource) -> Result<()> {
+        let _ = resource.resource.take();
+        if let Some(allocation) = resource.allocation.take() {
             self.free(allocation)
         } else {
-            // Committed resources are implicitly dropped when their refcount reaches 0.
-            // We only have to change the allocation and size tracking.
+            // Dx12 Committed resources are implicitly dropped when their refcount reaches 0.
+            // We only have to update the tracked allocation count and memory usage.
             if let Some(memory_type_index) = resource.memory_type_index {
                 let memory_type = &mut self.memory_types[memory_type_index];
 
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -17,14 +17,14 @@ use crate::{
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum AllocationScheme {
-    /// This allocation will be for a buffer, and it will be driver managed using vulkans DedicatedAllocation feature.
-    /// The driver will be able to perform optimizations in some cases with this type of allocation.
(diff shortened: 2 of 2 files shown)


src/d3d12/mod.rs
```diff
@@ -258,12 +258,26 @@ pub enum ResourceType<'a> {
...
+impl Resource {
+    pub fn resource(&self) -> &ID3D12Resource {
+        self.resource.as_ref().expect("Resource was already freed.")
+    }
+}
+
+impl Drop for Resource {
+    fn drop(&mut self) {
+        if self.resource.is_some() {
+            warn!("Dropping resource that was not freed. Was `Allocator::free_resource()` not called?")
```
MarijnS95, line 276: Nit: this means that the whole resource (not just the allocation) was not freed?

src/d3d12/mod.rs
```diff
@@ -903,13 +917,13 @@ impl Allocator {
...
-        let _ = ManuallyDrop::into_inner(resource.resource);
-        if let Some(allocation) = resource.allocation {
+    pub fn free_resource(&mut self, mut resource: Resource) -> Result<()> {
+        let _ = resource.resource.take();
+        if let Some(allocation) = resource.allocation.take() {
             self.free(allocation)
         } else {
-            // Committed resources are implicitly dropped when their refcount reaches 0.
-            // We only have to change the allocation and size tracking.
+            // Dx12 Committed resources are implicitly dropped when their refcount reaches 0.
```
MarijnS95, line 925: As with the previous review, I think you can leave this line out and instead add a comment above `resource.take()` to clarify that the resource has to be dropped _before_ the allocation. Add a note that `ID3D12Resource` is a COM object that performs refcounting internally, so we can never be 100% sure that it is being dropped on the spot...

That is, unless windows-rs exposes the refcounting result from `.Release()`, I'll create an issue for that since it allows us to assert/warn when the returned value is not `0`.

src/vulkan/mod.rs
```diff
@@ -17,14 +17,14 @@ use crate::{
...
-    /// This allocation will be managed by the GPU allocator.
-    /// It is possible that the allocation will reside in an allocation block shared with other resources.
+    /// Perform a dedicated, driver-managed allocation for the given buffer, allowing
+    /// it to perform optimizations on this type of allocation.
+    DedicatedBuffer(vk::Buffer),
+    /// Perform a dedicated, driver-managed allocation for the given image, allowing
+    /// it to perform optimizations on this type of allocation.
+    DedicatedImage(vk::Image),
+    /// The memory for this resource will be allocated and managed by gpu-allocator.
+    /// It is possible that the allocation will reside in memory shared with other resources.
```
MarijnS95, line 27:
```suggestion
    /// It is possible that the allocation will reside in [`vk::Memory`] shared with other resources.
```

We have `vk` in scope right? Otherwise "allocation block" is more clear :)

src/d3d12/mod.rs
```diff
@@ -903,13 +917,13 @@ impl Allocator {
     }
 
     /// Free a resource and its memory.
-    pub fn free_resource(&mut self, resource: Resource) -> Result<()> {
-        let _ = ManuallyDrop::into_inner(resource.resource);
-        if let Some(allocation) = resource.allocation {
+    pub fn free_resource(&mut self, mut resource: Resource) -> Result<()> {
+        let _ = resource.resource.take();
```
MarijnS95, line 921: Make this a `.take().expect("Resource was already freed.")`

src/d3d12/mod.rs
```diff
@@ -903,13 +917,13 @@ impl Allocator {
...
+        if let Some(allocation) = resource.allocation.take() {
```
MarijnS95, line 922: Never mind about this `.take()`, the field is implicitly moved and the compiler disallows you from using it entirely (instead of treating it as a possibly-`None` value:

```
borrow of partially moved value: `resource.allocation`
partial move occurs because value has type `d3d12::Allocation`, which does not implement the `Copy` trait
```

```suggestion
        if let Some(allocation) = resource.allocation {
```

MarijnS95: Ah, you can't because of the custom `Drop` implementation, never mind!

src/d3d12/mod.rs
MarijnS95: That said, should we add some debug-printing context here so that user knows _which_ allocation it was (do we have `allocation.name`)?

janie177: Unfortunately there's not always an allocation (in the case this was a CommittedResource). I can take a look if we can get more information from the resource pointer, because I think dx12 lets you set a custom name. But that might have been elsewhere.

src/d3d12/mod.rs
janie177: I don't think that's possible because the function consumes the `Resource` object. Since the field is private, I don't think it can ever be freed twice.

src/vulkan/mod.rs
janie177: Clippy was complaining that there was no `Memory` defined in `vk`. I suppose allocation block works too but I have no strong preference towards either.

src/d3d12/mod.rs
MarijnS95: Anything in our implementation can accidentally `.take()` it later down the line, an explicit `.expect()` makes this more clear :)

src/d3d12/mod.rs
MarijnS95: Ah yes, I unmarked this as resolved because of the second comment, let me strike through anything that is irrelevant after re-reading :)

src/vulkan/mod.rs
MarijnS95: `DeviceMemory` maybe? My mind might be fuzzy on the exact type naming.

src/d3d12/mod.rs
janie177: That is fair, I have adjusted the comment to at least mention that it's about `CommittedResource` in that branch. I moved part of the comment and clarified it a bit 👍

janie177:
> One more request: since you inserted the new `Resource` struct etc right in beween `struct Allocation` and `unsafe impl Send/Sync for Allocation`, can you move them back together; or better yet, remove them entirely? windows-rs implements these for COM, and all the other fields should have it too (but please test that!).
> 
> The whole idea of these marker threads is that they are auto-derived... and when they aren't Rust has a good reason to not do so. `unsafe impl` should be used extremely sparingly when we can justify that it is safe to do so...

I have moved them back together, but didn't remove the unsafe impls as I don't have much time to properly test stuff now. Feel free to take over this branch if you wish to do that in this PR though!

src/d3d12/mod.rs
janie177: Fair enough, I'll add it.

MarijnS95:
I've opened a more rigorous PR to tackle this! #152

src/d3d12/mod.rs
MarijnS95: Asked in https://github.com/microsoft/windows-rs/issues/2291, for us it's always `0` with the provided test fortunately :)

PR #161, opened 2023-03-24
Configurable allocation sizes

### LINKED ISSUE #156: Configurable allocation size
Currently it seems like gpu-allocator allocated a 256MB buffer on the gpu every time you initialize it. `wgpu` isn't only used for games, so it'd be nice if we could configure the block size to be smaller for eg. apps like [wezterm](https://github.com/wez/wezterm.git).

Elabajaba:
This adds an `AllocationSizes` struct to both the Vulkan and DX12 `Allocator` and `AllocatorCreateDesc` which allows users to configure the size of the memory blocks that are created to be suballocated from.

This doesn't currently clamp it to 256MB, which afaik is a problem for systems that don't support Resizable BAR, but I also don't know how to properly detect if someone is or isn't using Resizable BAR.

I also don't know if we need to enforce alignment, and if so what values do we need to use for alignment?

Closes #156

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ let mut allocator = Allocator::new(&AllocatorCreateDesc {
     physical_device,
     debug_settings: Default::default(),
     buffer_device_address: true,  // Ideally, check the BufferDeviceAddressFeatures struct.
+    allocation_sizes: Default::default(),
 });
 ```
 
@@ -78,6 +79,7 @@ use gpu_allocator::d3d12::*;
 let mut allocator = Allocator::new(&AllocatorCreateDesc {
     device,
     debug_settings: Default::default(),
+    allocation_sizes: Default::default(),
 });
 ```
 
--- a/examples/d3d12-buffer-winrs.rs
+++ b/examples/d3d12-buffer-winrs.rs
@@ -90,6 +90,7 @@ fn main() -> Result<()> {
     let mut allocator = Allocator::new(&AllocatorCreateDesc {
         device: device.clone(),
         debug_settings: Default::default(),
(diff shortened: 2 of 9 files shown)


manon-traverse (approved):
The 256MB block was just value we picked and worked for us since. Being able to configure this looks like a good change ^.^

Elabajaba:
I think it needs to be 4MB aligned for DX12 MSAA support? https://github.com/Traverse-Research/gpu-allocator/blob/5ff4300b3ac47c939aac3da2934c31e771533aec/src/d3d12/mod.rs#L338

Would you rather I clamp it between 4MB and 256MB and silently force 4MB alignment, or return an error on creation if the values the user passes in aren't aligned?

Elabajaba:
Afaik for vulkan it'd technically be more correct to move the alignment requirement check to when the allocator is created and query the device for `minMemoryMapAlignment`, but based on https://vulkan.gpuinfo.org/displaydevicelimit.php?name=minMemoryMapAlignment I don't think it's really necessary.

Elabajaba:
Bumping this, @manon-traverse @MarijnS95 is what I have here with just allowing setting a static allocation size within a reasonable range acceptable?

If you'd like, I could move it into the backend impls and query the device for minimum/maximum allocation sizes, but I don't think it's necessary with the  4MB to 256MB clamp I'm currently using.

We could also try to do some fancy automatically scaling heap size based on memory usage, but that seems a lot more complicated and may have performance pitfalls.

manon-traverse:
Sorry for not paying attention to this PR!
I think it's fine the way you have it right now, and I'll merge it

PR #167, opened 2023-05-10
Use `PhantomData` in `MappedAllocationSlab` and fix merge

fu5ha:
Small followup to #138 applying a slight optimisation and fixing a bad auto-merge.


fu5ha:
:hmm: not sure what the readme step is but it does seem to be failing 😅

PR #174, opened 2023-08-26
Fix minimal versions

SkiFire13:
I was trying to compile `bevy` using [`cargo minimal-versions`](https://github.com/taiki-e/cargo-minimal-versions) and noticed that `gpu-allocator` depended on `backtrace 0.3` and `log 0.4`, but actually used features from `backtrace 0.3.3` (`Backtrace::new_unresolved` and `Backtrace::resolve`) and `log 0.4.9` (didn't import the `log!` macro when calling `log::warn!`; could be fixed by importing it too if you prefer it, but looks unnecessary), which made the build fail.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ include = [
 all-features = true
 
 [dependencies]
-log = "0.4"
+log = "0.4.9"
 thiserror = "1.0"
 presser = { version = "0.3" }
 # Only needed for vulkan.  Disable all default features as good practice,


MarijnS95:
Would you mind contributing a `cargo +nightly generate-lockfile -Zminimal-versions` etc workflow to our CI so that this problem can be prevented in the future?

SkiFire13:
I wound't mind, but as is it would not pass due to problems in transitive dependencies, in particular the minimal version for `winapi` ends up being `0.2.5` which doesn't compile (I'm not sure what should be changed to fix this. `0.2.7` is the minimum where it compiles on my machine but it feels wrong to just add it to this crate) and `imgui-sys` should depend on `cc 1.0.3` rather than `cc 1.0` (I opened imgui-rs/imgui-rs#744 for that).

The CI workflow should be something like this:

```yml
  minimal-versions:
    name: Check minimal versions
    strategy:
      matrix:
        include:
        - os: ubuntu-latest
          features: vulkan,visualizer
        - os: windows-latest
          features: vulkan,visualizer,d3d12,public-winapi
    runs-on: ${{ matrix.os }}
    steps:
      - uses: actions/checkout@v3
      - uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: nightly
          override: true
      - uses: taiki-e/install-action@cargo-hack
      - uses: taiki-e/install-action@cargo-minimal-versions
      - run: cargo minimal-versions build --workspace --features ${{ matrix.features }} --no-default-features
```

MarijnS95:
> and also ran into a failed build due to this `backtrace` version `0.3.0` mis-dependency.

The backtrace crate is now entirely gone: #186 :tada:

PR #176, opened 2023-09-21
🪛 Replace `imgui` with `egui`

tosti007:
This PR replaces `imgui` with `egui` for the visualizations. 

Some notes:
1. I did try to make the code a bit cleaner along the way, but I think the visualization code can still use some de-duplication.
2. I removed the examples, as keeping them would mean having to rewrite the renderers to `egui` as well, we can add these again later is desired.
3. I kept the functionality as close as possible to the original, but I could not get the "focus" feature to work.
4. The memory chunk visualization uses the optimized scroll area logic, such that only the visible lines are drawn (this is a must for large blocks, otherwise FPS drops heavily).

image


Jasper-Bekkers (approved):
Happy to see this change, having egui integration for us is significantly more important and imgui was holding us back a bit. 

Unfortunate that we're losing some of the samples, but if somebody needs them they're free to port them over to egui.

fu5ha:
Nice :D

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "rust-analyzer.cargo.features": [
+        "visualizer"
```
Jasper-Bekkers, line 3: I don't think we should check this in

Jasper-Bekkers (commented):
I think the readme still needs some work because CI is complaining, but after that we can merge this and make a release

.vscode/settings.json
tosti007: My bad! This made development easier and I accidentally checked it in.

src/allocator/mod.rs
```diff
@@ -16,10 +16,22 @@ pub(crate) enum AllocationType {
...
+    #[allow(dead_code)] // dead_code allowed as this is used in for the visualizer feature
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Free => "Free",
+            Self::Linear => "Linear",
+            Self::NonLinear => "Non-Linear",
+        }
+    }
+}
+
```
manon-traverse, line 29: Perhaps this dead code could also be marked as conditional compilation under the visualizer feature? Not sure how big of a mess that will result in.

tosti007: Yea that'd work too, it probably won't be much of a difference, but I wouldn't allow other users of the lib to use this function without also using the `visualizer` feature. Right now (imgui) the codebase is kinda taking both approaches.

tosti007: Oh you're right that this type is `pub(crate)`, I'll change it to conditional compilation where possible.

manon-traverse (approved):

PR #184, opened 2023-10-12
Use std::backtrace

stefnotch:
Supersedes https://github.com/Traverse-Research/gpu-allocator/pull/183 . I picked `Option<Arc<Backtrace>>`, which shouldn't have any overhead, as long as the backtrace flag is not set.

I also bumped up the Rust edition, since the backtrace was added in Rust 1.65.0, which has been released well after the Rust 2021 edition has been released.

### DIFF
--- a/examples/vulkan-buffer-backtrace.rs
+++ b/examples/vulkan-buffer-backtrace.rs
@@ -1,0 +1,213 @@
+use ash::vk;
+use log::info;
+
+use std::default::Default;
+use std::ffi::CString;
+
+use gpu_allocator::vulkan::{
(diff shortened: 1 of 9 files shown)


examples/vulkan-buffer-backtrace.rs
stefnotch, line 1: I can also remove this file again, it was mainly for testing the backtrace feature.

On that note, I'd love it if there were a visualizer example that displayed backtraces.

Jasper-Bekkers (approved):

stefnotch:
Closing in favour of #186

PR #186, opened 2023-11-02
Replace `backtrace` crate with stabilized `std::backtrace` implementation

MarijnS95:
Sticking to [my promise](https://github.com/Traverse-Research/gpu-allocator/pull/183#issuecomment-1759302163) of updating and submitting this old branch in #183.

Fixes #183 (by no longer using the `backrace`/`cc` crate);
Fixes #184 (the old approach from that branch).

---

Rust 1.65 [stabilized `std::backtrace::Backtrace`], which we can use in place of the `backtrace` crate to reduce our dependency stack.

Normally `Backtrace::capture()` would listen to the `RUST_BACKTRACE` and `RUST_LIB_BACKTRACE` environment variables, but this is unsuitable for us as capturing backtraces is configured via boolean debug feature flags in the `AllocatorDebugSettings` struct.  Fortunately `Backtrace::force_capture()` exists which circumvents these env var checks and always returns a backtrace, and is hence used in the codebase here.

Unfortuantely the type no longer implements `Clone` like `backtrace::Backtrace`, requiring us to wrap it in an `Arc` (because most of our types are thread-safe) to clone the `Backtrace` around various (sub)allocations and statistics reports.

[stabilized `std::backtrace::Backtrace`]: https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#stabilized-apis

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ homepage = "https://github.com/Traverse-Research/gpu-allocator"
 repository = "https://github.com/Traverse-Research/gpu-allocator"
 keywords = ["vulkan", "memory", "allocator"]
 documentation = "https://docs.rs/gpu-allocator/"
+rust-version = "1.65"
 
 include = [
     "/README.md",
@@ -19,7 +20,6 @@ include = [
 ]
 
 [dependencies]
-backtrace = "0.3"
 log = "0.4"
 thiserror = "1.0"
 presser = { version = "0.3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "gpu-allocator"
 version = "0.24.0"
 authors = ["Traverse Research <opensource@traverseresearch.nl>"]
-edition = "2018"
+edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "Memory allocator for GPU memory in Vulkan and DirectX 12"
 categories = ["rendering", "rendering::graphics-api"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -78,7 +78,6 @@ rustflags = [
     # END - Embark standard lints v6 for Rust 1.55+
 
     # Our additional lints
-    "-Wclippy::clone_on_ref_ptr",
     "-Wclippy::cognitive_complexity",
     "-Wclippy::needless_pass_by_value",
     "-Wclippy::option_if_let_else",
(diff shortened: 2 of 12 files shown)


Jasper-Bekkers:
I'm not comfortable having the stack trace collection of this crate be dependent on a rust specific environment variable. This is a feature of this crate and should thus be controlled by us / users of this crate directly from code.

MarijnS95:
@Jasper-Bekkers the comment exactly explains that that is the reason that `Backtrace::force_capture()` is used over `Backtrace::capture()`, because the former _ignores Rust-specific environment variables_. I see now that that comment wasn't super clear, so I've completely rewritten that paragraph.

Cargo.toml
```diff
@@ -10,6 +10,7 @@ homepage = "https://github.com/Traverse-Research/gpu-allocator"
 repository = "https://github.com/Traverse-Research/gpu-allocator"
 keywords = ["vulkan", "memory", "allocator"]
 documentation = "https://docs.rs/gpu-allocator/"
+rust-version = "1.65"
```
stefnotch, line 13: Would it make sense to instead bump the Rust edition? Or do both?

After all, Rust version 1.65 already supports the 2021 edition.

stefnotch (approved):
Awesome, thank you for this proper migration to Rust's backtrace. I'll definitely celebrate having slightly fewer dependencies in my apps :)

Cargo.toml
MarijnS95: We definitely could, but not _instead_, as edition 2021 implies `1.56.0` (https://doc.rust-lang.org/edition-guide/rust-2021/index.html) while this feature was only stabilized in `1.65` (9 releases later).

Jasper-Bekkers: @MarijnS95 Could you update the readme to mention this msrv as well?

MarijnS95: Added a CI check to enforce our MSRV which is even more important IMO. Unfortunately "bumbing" `egui` in https://github.com/Traverse-Research/gpu-allocator/pull/185 raised our MSRV mich higher than this, all the way to 1.70.

Will think about how to solve this: switch to 1.70 or not test all the features?

MarijnS95: @Jasper-Bekkers I ended up removing `visualizer` from the CI MSRV test, and documented in `README` what is _and what is not_ included in our MSRV :)

Jasper-Bekkers (approved):

PR #187, opened 2023-11-03
CI: Replace unnecessary `actions-rs` steps with direct `run: cargo ...`

MarijnS95:
GitHub Actions' `runner-images` already come fully loaded with all the Rust tools that we need [1]: remove the bloated and outdated `actions-rs/toolchain` setup step for that, and also remove the `actions-rs/cargo` action that only serves as a more convoluted way to run simple `cargo` shell commands.

Besides, `actions-rs` are unmaintained and outdated, and throw an alarming number of spammy warnings in CI build output.

[1]: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#rust-tools

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,15 +15,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


Jasper-Bekkers (approved):
Much better.

PR #199, opened 2024-01-29
🍎 Add Metal implementation

FlannyH:
Implements gpu-allocator for the Metal API. This PR does not implement the visualizer, as I'm currently unable to test that on Metal

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,12 @@ ash = { version = ">=0.34, <=0.37", optional = true, default-features = false, f
 egui = { version = ">=0.24, <=0.26", optional = true, default-features = false }
 egui_extras = { version = ">=0.24, <=0.26", optional = true, default-features = false }
 
+[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
+metal = { version = "0.27.0", default-features = false, features = [
+    "link",
+    "dispatch",
+] }
+
 [target.'cfg(windows)'.dependencies]
 # Only needed for public-winapi interop helpers
 winapi = { version = "0.3.9", features = ["d3d12", "winerror", "impl-default", "impl-debug"], optional = true }
@@ -88,3 +94,6 @@ d3d12 = ["windows"]
 public-winapi = ["dep:winapi"]
 
 default = ["d3d12", "vulkan"]
+
+[patch.crates-io]
+metal = { git = "https://github.com/Traverse-Research/metal-rs", rev = "d40f71a" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,9 @@ thiserror = "1.0"
 presser = { version = "0.3" }
 # Only needed for vulkan.  Disable all default features as good practice,
 # such as the ability to link/load a Vulkan library.
-ash = { version = ">=0.34, <=0.37", optional = true, default-features = false, features = ["debug"] }
+ash = { version = ">=0.34, <=0.37", optional = true, default-features = false, features = [
+    "debug",
+] }
 # Only needed for visualizer.
 egui = { version = ">=0.24, <=0.26", optional = true, default-features = false }
 egui_extras = { version = ">=0.24, <=0.26", optional = true, default-features = false }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -28,14 +23,27 @@ thiserror = "1.0"
 presser = { version = "0.3" }
 # Only needed for vulkan.  Disable all default features as good practice,
 # such as the ability to link/load a Vulkan library.
-ash = { version = ">=0.34, <=0.37", optional = true, default-features = false, features = ["debug"] }
+ash = { version = ">=0.34, <=0.37", optional = true, default-features = false, features = [
+    "debug",
+] }
 # Only needed for visualizer.
```
Jasper-Bekkers, line 29: @FlannyH Please don't reformat Cargo.toml :-)

src/metal/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+
+impl MemoryType {
+    fn allocate(
+        &mut self,
+        device: Arc<metal::Device>,
+        desc: &AllocationCreateDesc<'_>,
+        backtrace: Arc<Backtrace>,
+        allocation_sizes: &AllocationSizes,
+    ) -> Result<Allocation> {
+        let allocation_type = AllocationType::Linear;
```
Jasper-Bekkers, line 206: I don't think this should be the default, not sure how this parameter is used but Linear is usually the least optimal choice.

src/metal/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+            1,
+            desc.name,
+            backtrace,
+        );
+        let (offset, chunk_id) = match allocation {
+            Err(AllocationError::OutOfMemory) => Err(AllocationError::Internal(
+                "Allocation that must succeed failed. This is a bug in the allocator.".into(),
+            )),
+            a => a,
+        }?;
```
Jasper-Bekkers, line 318: I don't quite understand how an OutOfMemory error is a bug in the allocator? 

Additionally, I think it's a good idea to return along with the error messages the name of the allocation we tried to do, though that's more of a general remark then a remark for this specific case alone.

Cargo.toml
FlannyH: Yeah that happened by accident after installing a plugin, I'll fix it in a sec

src/metal/mod.rs
FlannyH: This is also in the Dx12 implementation, which I used as a base to create the Metal implementation, so this might be something to look at in the Dx12 implementation as well then

Cargo.toml
FlannyH: Fixed in latest commit

src/metal/mod.rs
FlannyH: This is the way it was done in Dx12 as well. Should we update both Dx12 and Metal to do it like Vulkan, which sets this to Linear or NonLinear based on the AllocationCreateDesc passed into the allocate function?

src/metal/mod.rs
FlannyH: It shouldn't be possible to get an OutOfMemory error here, because in this part of the function, we're allocating from a newly created memory block, which should be empty and have enough space for the allocation

manon-traverse: The allocation must succeed since it just managed to allocate a new empty block that is large enough for the allocation.

src/metal/mod.rs
manon-traverse: DX12 doesn't make a distinction between linear and non-linear resources, while Vulkan does.

@FlannyH you should look up in the Metal docs if there are any restrictions on resource allocation. Mainly if textures and buffers can live the same memory blocks or if they need padding between them.

FlannyH: I'll have a look

src/lib.rs
```diff
@@ -170,6 +170,10 @@ pub mod vulkan;
 #[cfg(all(windows, feature = "d3d12"))]
 pub mod d3d12;
 
+// todo(lily): add a feature toggle for metal?
+#[cfg(any(target_os = "macos", target_os = "ios"))]
```
manon-traverse, line 174:
```suggestion
#[cfg(all(any(target_os = "macos", target_os = "ios"), feature = "metal"))]
```

This should resolve your TODO comment.

FlannyH: Changed it in latest commit

src/metal/mod.rs
manon-traverse: I can't find anything in the metal docs either, so I think it should be fine without it.

manon-traverse (approved):

PR #207, opened 2024-02-21
👨‍🔬Total memory in Allocation Breakdown

Jasper-Bekkers:
We've had this feature previously but it got nerf'd during our transition from egui to imgui. It's small but super useful to be able to see the total amount of memory under various filters (or just the total when no filter is applied).

### DIFF
--- a/src/visualizer/allocation_reports.rs
+++ b/src/visualizer/allocation_reports.rs
@@ -37,9 +37,13 @@ pub(crate) fn render_allocation_reports_ui(
         .enumerate()
         .filter(|(_, report)| report.name.to_lowercase().contains(&breakdown_filter))
         .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


src/visualizer/allocation_reports.rs
```diff
@@ -37,9 +37,13 @@ pub(crate) fn render_allocation_reports_ui(
         .enumerate()
         .filter(|(_, report)| report.name.to_lowercase().contains(&breakdown_filter))
         .collect::<Vec<_>>();
+    let total_size_under_filter: u64 = allocations.iter().map(|a| a.1.size).sum();
 
-    let row_height = ui.text_style_height(&egui::TextStyle::Body);
+    ui.horizontal(|ui| {
```
tosti007, line 42: Nit: there's only a single label in here, so the horizontal isn't needed.

tosti007 (approved):

PR #213, opened 2024-03-11
cargo: Bump `windows` crate range to `0.53-0.57`

MarijnS95:
Allow `windows 0.53` all the way up to and including `0.57` to be used, and remove some unused features from the core crate.


Systemcluster:
The `ComInterface` trait seems to be just `Interface` since `0.53`: https://github.com/microsoft/windows-rs/issues/2875#issuecomment-1962332067

I'm not sure why it's not mentioned in the release notes. `Interface` seems to work down to at least `0.51` though, so I think you can just replace it without bumping the lower version bound.

cwfitzgerald:
This would be great to get in, I've started depending on it locally.

MarijnS95:
Looks like `0.57` requires an MSRV bump to Rust 1.70.

tosti007 (approved):

MarijnS95:
Ugh, the merge-commit title failed to update.

PR #214, opened 2024-03-21
Make `metal` dependency rely on the `metal` feature

MarijnS95:
We seem to have this weird case where the `metal` crate is always included (on supported target OSes), even if the non-default `metal` feature isn't turned on.  Found this while wondering why `cargo` wasn't complaining about a clash thanks to the implicit feature created by an optional crate with an identically-named feature in the `[features]` table without [the `dep:` rename prefix], turns out that the `metal` crate wasn't `optional = true` in the first place.  Fix that.

[the `dep:` rename prefix]: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ all-features = true
 log = "0.4"
 thiserror = "1.0"
 presser = { version = "0.3" }
-# Only needed for vulkan.  Disable all default features as good practice,
+# Only needed for Vulkan.  Disable all default features as good practice,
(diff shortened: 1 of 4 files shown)


MarijnS95:
Just in case, since this also touches up some typical clippy lints that I haven't been able to (build)test, let's wait for and rebase on #215 once the metal backend is properly tested in the CI.

FlannyH (approved):

PR #215, opened 2024-03-21
CI: Also test and lint Metal (MacOS/iOS) code

MarijnS95:


MarijnS95:
This all is really terrible as `metal-rs` is depending on ancient **unmaintained** mac crates that haven't been touched _over four years_.

Really hoping the upstream migration to `objc2` is being pushed through sooner than later.

FlannyH (approved):

PR #221, opened 2024-04-19
d3d12: Add support for castable formats on resource allocations

EmilioLaiso:
This feature has a dependency on `ID3D12Device12` due to the need of using `GetResourceAllocationInfo3`.
The PR is a draft because I'm still figuring out castable formats support on VK and Metal. 
From early research it looks like VK won't require any change in gpu-allocator as it's mainly achieved through a combination of `VK_IMAGE_CREATE_` bitflags which are outside of the scope of the allocator and handled on application side.

Metal is still being figured out.

### DIFF
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -245,7 +245,8 @@ pub enum ID3D12DeviceVersion {
     Device(ID3D12Device),
     /// Required for enhanced barrier support, i.e. when using
     /// [`ResourceStateOrBarrierLayout::BarrierLayout`].
-    /// As well as for castable formats support.
+    Device10(ID3D12Device10),
+    /// Required for castable formats support
     Device12(ID3D12Device12),
 }
 
@@ -255,7 +256,8 @@ impl std::ops::Deref for ID3D12DeviceVersion {
     fn deref(&self) -> &Self::Target {
         match self {
             Self::Device(device) => device,
-            // Windows-rs hides CanInto, we know that Device12 is a subclass of Device but there's not even a Deref.
+            // Windows-rs hides CanInto, we know that Device10/Device12 is a subclass of Device but there's not even a Deref.
+            Self::Device10(device10) => windows::core::CanInto::can_into(device10),
             Self::Device12(device12) => windows::core::CanInto::can_into(device12),
         }
     }
@@ -811,6 +813,26 @@ impl Allocator {
         }
     }
 
+    fn d3d12_resource_desc_1(
+        desc: &D3D12_RESOURCE_DESC,
+        sampler_feedback_mip_region: D3D12_MIP_REGION,
+    ) -> D3D12_RESOURCE_DESC1 {
+        D3D12_RESOURCE_DESC1 {
+            Dimension: desc.Dimension,
+            Alignment: desc.Alignment,
+            Width: desc.Width,
+            Height: desc.Height,
+            DepthOrArraySize: desc.DepthOrArraySize,
+            MipLevels: desc.MipLevels,
+            Format: desc.Format,
+            SampleDesc: desc.SampleDesc,
+            Layout: desc.Layout,
+            Flags: desc.Flags,
+            // TODO: This is the only new field
+            SamplerFeedbackMipRegion: sampler_feedback_mip_region,
+        }
+    }
+
     /// Create a resource according to the provided parameters.
     /// Created resources should be freed at the end of their lifetime by calling [`Self::free_resource()`].
     pub fn create_resource(&mut self, desc: &ResourceCreateDesc<'_>) -> Result<Resource> {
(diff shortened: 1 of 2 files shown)


src/d3d12/mod.rs
```diff
@@ -242,6 +246,8 @@ pub enum ID3D12DeviceVersion {
     /// Required for enhanced barrier support, i.e. when using
     /// [`ResourceStateOrBarrierLayout::BarrierLayout`].
     Device10(ID3D12Device10),
+    /// Required for castable formats support
```
MarijnS95, line 249: Mention that this is in addition to `Device10`?

src/d3d12/mod.rs
```diff
@@ -806,6 +813,26 @@ impl Allocator {
...
+            Width: desc.Width,
+            Height: desc.Height,
+            DepthOrArraySize: desc.DepthOrArraySize,
+            MipLevels: desc.MipLevels,
+            Format: desc.Format,
+            SampleDesc: desc.SampleDesc,
+            Layout: desc.Layout,
+            Flags: desc.Flags,
+            // TODO: This is the only new field
+            SamplerFeedbackMipRegion: sampler_feedback_mip_region,
```
MarijnS95, line 832: What do you want to do here?

src/d3d12/mod.rs
```diff
@@ -959,7 +1078,35 @@ impl Allocator {
...
-                                None, // TODO: https://github.com/microsoft/DirectX-Specs/blob/master/d3d/VulkanOn12.md#format-list-casting
+                                None,
+                                &mut result,
+                            )
+                        }
+                        (
+                            ID3D12DeviceVersion::Device12(device),
+                            ResourceStateOrBarrierLayout::BarrierLayout(initial_layout),
+                        ) => {
+                            let resource_desc1 = D3D12_RESOURCE_DESC1 {
```
MarijnS95, line 1089: You added a helper for this no?

src/d3d12/mod.rs
```diff
@@ -822,6 +849,10 @@ impl Allocator {
                 if let Err(e) = unsafe {
                     match (&self.device, desc.initial_state_or_layout) {
                         (device, ResourceStateOrBarrierLayout::ResourceState(initial_state)) => {
+                            if !desc.castable_formats.is_empty() {
+                                warn!("Requesting castable formats, but this feature requires ID3D12Device12");
+                            }
```
MarijnS95, line 854: Given the match pattern this is not true as written; one can always create a `gpu-allocator` with a `Device12` but then use the legacy `ResourceState`, those two things are specifically disconnected.

src/d3d12/mod.rs
```diff
@@ -835,20 +866,30 @@ impl Allocator {
...
+                            );
+
+                            device.CreateCommittedResource3(
+                                *heap_properties,
+                                *heap_flags,
+                                &resource_desc1,
+                                initial_layout,
+                                clear_value,
+                                None, // TODO
+                                None,
```
MarijnS95, line 881: That's the castable stuff, no?

src/d3d12/mod.rs
MarijnS95: However, it seems that the only function that takes the castable formats as argument requires the new enhanced-barrier layout and can no longer be used with legacy resource states?

MarijnS95: In which case I think we should flip the `match` around to make it more natural as written.

src/d3d12/mod.rs
```diff
@@ -872,9 +913,46 @@ impl Allocator {
...
-                let allocation_info = unsafe {
-                    self.device
-                        .GetResourceAllocationInfo(0, &[*desc.resource_desc])
+                let allocation_info = match &self.device {
+                    ID3D12DeviceVersion::Device(device) => unsafe {
+                        device.GetResourceAllocationInfo(0, &[*desc.resource_desc])
+                    },
+                    ID3D12DeviceVersion::Device10(device) => unsafe {
+                        device.GetResourceAllocationInfo(0, &[*desc.resource_desc])
+                    },
```
MarijnS95, line 922: Clippy should suggest to merge identical arms?

src/d3d12/mod.rs
```diff
@@ -906,9 +984,46 @@ impl Allocator {
...
+                            device.GetResourceAllocationInfo3(
+                                0,
+                                resource_descs.len() as u32,
+                                resource_descs.as_ptr(),
+                                num_castable_formats_opt,
+                                castable_formats_opt,
+                                None,
+                            )
+                        },
                     };
```
MarijnS95, line 1027: Can we deduplicate this whole thing with the above?

src/d3d12/mod.rs
EmilioLaiso: Can't because device is of different types

src/d3d12/mod.rs
EmilioLaiso: Yes but called on `Device10`, so we don't add the castable formats in that code path.

src/d3d12/mod.rs
EmilioLaiso: What do you mean?

src/d3d12/mod.rs
MarijnS95: Not having `Deref` here is a bug that will be solved on a future version of windows-rs, unfortunately.

src/d3d12/mod.rs
MarijnS95: I thought this was new code. Perhaps it belongs on where the function is called with a "default" mip region?

MarijnS95: Fwiw with flipping I don't mean `(A, B) =>` -> `(B, A) =>` because that has no effect. But rather something like:

```rust
match (d, s) {
(device, ResourceState) if !desc.castable_formats.is_empty() => Err(Error::CastableFormatsRequiresEnhancedBarriers),
(Device12, EnhancedBarrier) => { your new stuff },
(device, EnhancedBarrier) if !desc.castable_formats.is_empty() => Err(Error::CastableFormatsRequiresatLeastDevice10),
(Device10, EnhancedBarrier) => { original code },
(device, EnhancedBarrier) => Err(AllocationError::BarrierLayoutNeedsDevice10), // Original code too
(device, ResourceState) => { original code for legacy barriers },
}
```

MarijnS95: Alternatively we could switch to runtime `QueryInterface` to do the checking more ad-hoc?

src/d3d12/mod.rs
MarijnS95: https://github.com/Traverse-Research/gpu-allocator/pull/221#discussion_r1572134935, looks like GitHub put it under the wrong thread.

src/d3d12/mod.rs
```diff
@@ -242,6 +246,8 @@ pub enum ID3D12DeviceVersion {
     /// Required for enhanced barrier support, i.e. when using
     /// [`ResourceStateOrBarrierLayout::BarrierLayout`].
     Device10(ID3D12Device10),
+    /// Required for castable formats support, requires the use of enhanced barrier
+    Device12(ID3D12Device12),
```
MarijnS95, line 250: More so meant to say that `Device12` _gives_ you enhanced barrier support.

The _requirement_ for EB to support castable formats is something that better fits on the function that takes this argument.

MarijnS95 (approved):
Reviewed in-person, implementation looks good besides one doc-comment!

(Also we should have had more doc comments on the functions that could be extended here but, since we don't, that's not up to you to fix)

MarijnS95 (approved):

PR #225, opened 2024-05-01
🤘 metal: Migrate to `objc2` architecture with `objc2-metal` bindings

MarijnS95:
The current `objc` crate stack is completely unmaintained and has severely fallen out of date with no updates for over 4 years.  The `metal-rs` crate, built on top of this architecture, is completely written by hand which is tedious to keep up-to-date, not to mention has inconsistencies in its implementation.

All of this is superseded by the new `objc2` crate stack built by @Madsmtm.  Beyond providing what seems like a better, safer abstraction over Objective-C, _all_ bindings are completely autogenerated meaning we'll no longer lag behind upstream bindings (requiring painstaking manual patching) or have inconsistencies in the implementations, as long as the generator is properly able to represent the bindings.

---

This PR is a draft to allow me to get a better understanding of Objective-C, as well as for @Madsmtm to chime in on the current use of Metal bindings.  Who also helpfully provided me [a branch with planned future improvements](https://github.com/madsmtm/objc2/tree/metal-wgpu), including changing the generated bindings to be closer to `metal-rs` to make other migrations easier to manage.

One important thing that is missing from this PR is interop with older `metal-rs` types.  `objc2(-metal)` has the typical `Id::from_raw()` interop that could allow us to upgrade `gpu-allocator` here while still using `metal-rs` for a little while longer in our own codebase.  Such interop could be provided like the `public-winapi` optional feature, or hand-rolled in our codebase if so desired.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,17 @@ egui = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 egui_extras = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-metal = { version = "0.29.0", default-features = false, features = ["link", "dispatch"], optional = true }
+objc2 = { version = "0.5.2", default-features = false, optional = true }
+objc2-foundation = { version = "0.2", default-features = false, optional = true }
+objc2-metal = { version = "0.2.1", default-features = false, features = [
+    "MTLAccelerationStructure",
+    "MTLBuffer",
+    "MTLDevice",
+    "MTLHeap",
+    "MTLResource",
+    "MTLTexture",
+    "std",
+], optional = true }
 
 [target.'cfg(windows)'.dependencies]
 # Only needed for public-winapi interop helpers
@@ -65,6 +75,11 @@ features = [
     "Win32_Graphics_Dxgi_Common",
 ]
 
+[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dev-dependencies]
+objc2-metal = { version = "0.2.1", default-features = false, features = [
+    "MTLPixelFormat",
+] }
+
 [[example]]
 name = "vulkan-buffer"
 required-features = ["vulkan", "ash/loaded"]
@@ -85,8 +100,8 @@ required-features = ["metal"]
 visualizer = ["dep:egui", "dep:egui_extras"]
 vulkan = ["dep:ash"]
 d3d12 = ["dep:windows"]
-metal = ["dep:metal"]
+metal = ["dep:objc2", "dep:objc2-metal", "dep:objc2-foundation"]
 # Expose helper functionality for winapi types to interface with gpu-allocator, which is primarily windows-rs driven
 public-winapi = ["dep:winapi"]
 
-default = ["d3d12", "vulkan"]
+default = ["d3d12", "vulkan", "metal"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ ash = { version = "0.38", optional = true, default-features = false, features = 
 egui = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 egui_extras = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 
-[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
+[target.'cfg(target_vendor = "apple")'.dependencies]
 objc2 = { version = "0.5.2", default-features = false, optional = true }
 objc2-foundation = { version = "0.2", default-features = false, optional = true }
 objc2-metal = { version = "0.2.1", default-features = false, features = [
@@ -75,7 +75,7 @@ features = [
     "Win32_Graphics_Dxgi_Common",
 ]
 
-[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dev-dependencies]
+[target.'cfg(target_vendor = "apple")'.dev-dependencies]
 objc2-metal = { version = "0.2.1", default-features = false, features = [
     "MTLPixelFormat",
 ] }
--- a/examples/metal-buffer.rs
+++ b/examples/metal-buffer.rs
@@ -1,12 +1,20 @@
-use std::sync::Arc;
-
 use gpu_allocator::metal::{AllocationCreateDesc, Allocator, AllocatorCreateDesc};
 use log::info;
+use metal::MTLDevice as _;
+use objc2::rc::Id;
+use objc2_foundation::NSArray;
+use objc2_metal as metal;
 
 fn main() {
     env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("trace")).init();
 
-    let device = Arc::new(metal::Device::system_default().unwrap());
+    // Allow the innards of objc2-metal to link the static function below:
+    // https://docs.rs/objc2-metal/0.2.2/objc2_metal/index.html
+    #[link(name = "CoreGraphics", kind = "framework")]
+    extern "C" {}
+
+    let device =
+        unsafe { Id::from_raw(metal::MTLCreateSystemDefaultDevice()) }.expect("No MTLDevice found");
 
     // Setting up the allocator
     let mut allocator = Allocator::new(&AllocatorCreateDesc {
@@ -60,11 +68,11 @@ fn main() {
 
     // Test allocating texture
     {
-        let texture_desc = metal::TextureDescriptor::new();
-        texture_desc.set_pixel_format(metal::MTLPixelFormat::RGBA8Unorm);
-        texture_desc.set_width(64);
-        texture_desc.set_height(64);
-        texture_desc.set_storage_mode(metal::MTLStorageMode::Private);
+        let texture_desc = unsafe { metal::MTLTextureDescriptor::new() };
+        texture_desc.setPixelFormat(metal::MTLPixelFormat::RGBA8Unorm);
+        unsafe { texture_desc.setWidth(64) };
+        unsafe { texture_desc.setHeight(64) };
+        texture_desc.setStorageMode(metal::MTLStorageMode::Private);
         let allocation_desc =
             AllocationCreateDesc::texture(&device, "Test allocation (Texture)", &texture_desc);
         let allocation = allocator.allocate(&allocation_desc).unwrap();
@@ -75,14 +83,14 @@ fn main() {
 
     // Test allocating acceleration structure
     {
-        let empty_array = metal::Array::from_slice(&[]);
-        let acc_desc = metal::PrimitiveAccelerationStructureDescriptor::descriptor();
-        acc_desc.set_geometry_descriptors(empty_array);
-        let sizes = device.acceleration_structure_sizes_with_descriptor(&acc_desc);
+        let empty_array = NSArray::from_slice(&[]);
+        let acc_desc = metal::MTLPrimitiveAccelerationStructureDescriptor::descriptor();
+        acc_desc.setGeometryDescriptors(Some(&empty_array));
+        let sizes = device.accelerationStructureSizesWithDescriptor(&acc_desc);
         let allocation_desc = AllocationCreateDesc::acceleration_structure_with_size(
             &device,
             "Test allocation (Acceleration structure)",
-            sizes.acceleration_structure_size,
+            sizes.accelerationStructureSize as u64,
             gpu_allocator::MemoryLocation::GpuOnly,
         );
         let allocation = allocator.allocate(&allocation_desc).unwrap();
--- a/src/metal/mod.rs
+++ b/src/metal/mod.rs
@@ -1,7 +1,10 @@
-#![deny(clippy::unimplemented, clippy::unwrap_used, clippy::ok_expect)]
 use std::{backtrace::Backtrace, sync::Arc};
 
 use log::debug;
+use metal::{MTLDevice as _, MTLHeap as _, MTLResource as _};
+use objc2::{rc::Retained, runtime::ProtocolObject};
+use objc2_foundation::NSString;
+use objc2_metal as metal;
 
 use crate::{
     allocator::{self, AllocatorReport, MemoryBlockReport},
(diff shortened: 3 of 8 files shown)


src/metal/mod.rs
```diff
@@ -80,63 +97,68 @@ pub struct AllocationCreateDesc<'a> {
...
     pub fn acceleration_structure_with_size(
-        device: &metal::Device,
+        device: &ProtocolObject<dyn metal::MTLDevice>,
         name: &'a str,
-        size: u64,
+        size: u64, // TODO: usize
         location: MemoryLocation,
     ) -> AllocationCreateDesc<'a> {
-        let size_and_align = device.heap_acceleration_structure_size_and_align_with_size(size);
+        // TODO: Why is this one unsafe?
```
madsmtm, line 143: Likely just because no-one has marked it as safe.

madsmtm (approved):

Cargo.toml
```diff
@@ -34,7 +34,18 @@ egui = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 egui_extras = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-metal = { version = "0.27.0", default-features = false, features = ["link", "dispatch"], optional = true }
+objc2 = { version = "0.5", default-features = false, optional = true }
+objc2-foundation = { version = "0.2", default-features = false, optional = true }
+objc2-metal = { version = "0.2.0", default-features = false, features = [
```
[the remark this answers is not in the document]
MarijnS95: Okay, I'll look into changing this. Turning off `default-features` is a bit of a habit after seeing that quite a few crates have overzealous/strange defaults. I'm rather a bit more explicit about what we turn on.

I wish `cargo`/`rustdoc` had better ways of documenting features (pages like https://docs.rs/crate/objc2/latest/features are currently meaningless) but I understand what these features do now.

---

> For example, the `objc2` crate has an `apple` feature flag which is enabled automatically by `objc2-metal`, simply because that doesn't _currently_ support GNUStep, but that might change in the future, and then this would stop compiling.

Does that also mean that these [runtime selection features](https://github.com/madsmtm/objc2/blob/5a891843844c69baf4c47f710f64ce59f2813249/crates/objc2/Cargo.toml#L94-L100) are mutually exclusive? If so that goes against what 

Since this is a library crate I don't want to make that choice for the user, and would like to turn off at much as possible in favour of the "end point" crate ultimately selecting the right thing.

src/metal/mod.rs
MarijnS95: Oh that's interesting. How much manual work is involved in improving the representation of the Metal API (or anything objc in general)?

I never want my implementations to be one way, and would love to aid in contributing back upstream bindings improvements. Is there some documentation or guidance in what goes into deducing that a function is safe (or any other thing that I might want to change), and how/where to ultimately mark it as such?

I'll look into your crate documentation meanwhile!

MarijnS95: https://github.com/madsmtm/objc2/blob/master/crates/header-translator/src/data/Metal.rs looks like this is where I'll want to start making changes.

Cargo.toml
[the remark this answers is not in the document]
madsmtm: Yup, the runtime features are mutually exclusive, don't know how to make that better.

In _theory_, you could turn off default features and expose your own `apple`/`gnustep` features, in practice I swear that you're not going to have a single user that uses GNUStep, so it'd just be wasted work.

I've been considering removing the features and making it dependent on the target triple instead, just to avoid this failure mode. It isn't the entire solution though, e.g. GNUStep would still require extra features per version, I think it can technically be used on Apple systems, and there also exist other runtimes like ObjFW, but it might still be better in practice.

src/metal/mod.rs
madsmtm: That is the place, yeah. I've been considering designing an "auto-safe" system where you could mark an entire class as being safe, and then all the methods on that class that take references and not e.g. pointers would be marked safe.

But it's a tough issue, and not a burden I wanted to tackle right now, especially in the political sense (I cannot really afford for `objc2` to be seen as something that plays quick and dirty with `unsafe`).

MarijnS95: I saw some of these methods were likely commented out and marked as unsafe because of possibly going out of bounds? E.g. `setWidth:` and friends, how do you deduce what happens here? And `heapAccelerationStructureSizeAndAlignWithSize:` too?

Cargo.toml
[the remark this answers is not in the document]
MarijnS95: The only alternative I know of is using manual `cfg`s that are painful to setup for consumer projects (`RUSTFLAGS=--cfg=apple`) without doing a feature passthrough or having a poor experience because of needing to always choose a runtime if it's not the default.

This recently came up with the `vello` crate as well, which is the bearer of all datastructures and logic but also has a default-enabled `wgpu` feature and dependency. Some support crates were not setting `default-features = false` and thus forcing every consumer to opt in to all those dependencies even if they ship their own renderer. The passthrough feature was decided on for the support crates, because they were reexporting the core `vello` crate in the root. It was disputed for a bit because the feature technically doesn't affect anything in the support crate otherwise.

src/metal/mod.rs
madsmtm: > how do you deduce what happens here

In reality, I don't really - the issue here is that I don't know enough about Metal to know what's safe and what isn't, the only reason that a lot of things are marked safe is mostly to retain approximate feature-compatibility with the `metal` crate, I haven't actually done a thorough review of everything there.

In the case of indexes I went with being overly cautious, they _looked_ like something that might very easily run into UB if you passed Metal some buffer and then set the size of the buffer to something else, but it's very probable that Metal protects against this internally, and that it isn't actually an issue.

Cargo.toml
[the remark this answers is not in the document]
madsmtm: I've deprecated the `"apple"` cargo feature in https://github.com/madsmtm/objc2/commit/9473e476a5dac2a97f4a53c675f689a82d331817 (will be released with `objc2 v0.5.2` / framework crates `v0.2.1`), in favour of using `target_vendor = "apple"` internally - so using `default-features = false, features = ["std"]` should work now.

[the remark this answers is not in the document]
MarijnS95: Thanks! I'll keep an eye out for that :)

src/metal/mod.rs
MarijnS95: That sounds very sensible. Sounds like the combination of common-sense and maybe runtime-testing (if I pass an invalid value: does Metal prevent me from (ab)using it by returning an error or otherwise)?

More importantly this might be something to document in the `header-translator` files directly (if not already done so), maybe serving a double purpose as `# Safety` docs when deducing that a function is explicitly unsafe (or maybe also why it would be "safe" if deduced as such).

---

> the only reason that a lot of things are marked safe is mostly to retain approximate feature-compatibility with the `metal` crate

Ouch; I hope any failures there to not mark things `unsafe` when they really are `unsafe` don't trickle down into your crates. I think that's the kind of compatibility you _shouldn't_ retain, to demonstrate that there's an obvious safety benefit in using `objc2.

Cargo.toml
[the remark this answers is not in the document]
MarijnS95: I see that the release was cut, and we can now rely on https://github.com/madsmtm/objc2/commit/d8db2f1352e3b5851cabd95799d10acfb8774138 :tada:

madsmtm (commented):
Glad to see that you're progressing here! I can't really help with the correctness of using the Metal API itself, am not experienced enough with it to do so, but I'll be happy to continue answering questions if you're using `objc2`-specific APIs that are confusing (this also goes for future PRs in this and other repos).

Cargo.toml
```diff
@@ -34,7 +34,17 @@ egui = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 egui_extras = { version = ">=0.24, <=0.27", optional = true, default-features = false }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
```
madsmtm, line 36: May I suggest `target_vendor = "apple"`, btw? If you support iOS, it's usually very little work to support tvOS and visionOS too.

MarijnS95: Done! Wasn't there a recent discussion about one of these being deprecated? Which cfg was it?

madsmtm: Tracking issue is [here](https://github.com/rust-lang/rust/issues/100343), though it won't be deprecated before a viable alternative (like `target_family = "apple"`) is available

examples/metal-buffer.rs
```diff
@@ -1,12 +1,16 @@
...
+use objc2::rc::Id;
+use objc2_foundation::NSArray;
+use objc2_metal as metal;
 
 fn main() {
     env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("trace")).init();
 
-    let device = Arc::new(metal::Device::system_default().unwrap());
+    let device = unsafe { metal::MTLCreateSystemDefaultDevice() };
+    // TODO: Not SendSync
```
Jasper-Bekkers, line 12: What does this todo mean?

Jasper-Bekkers (approved):

examples/metal-buffer.rs
MarijnS95: If I remember correctly, the new Rust clippy lint that complains about using a not-`Send`-`Sync` datastructure inside an `Arc` (since it can't be shared across threads, it's useless and an `Rc` should be used instead). This is not something to affect `gpu-allocator`, but will likely affect our framework.

examples/metal-buffer.rs
```diff
@@ -1,12 +1,16 @@
...
+use metal::MTLDevice as _;
+use objc2::rc::Id;
+use objc2_foundation::NSArray;
+use objc2_metal as metal;
 
 fn main() {
     env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("trace")).init();
 
-    let device = Arc::new(metal::Device::system_default().unwrap());
+    let device = unsafe { metal::MTLCreateSystemDefaultDevice() };
```
MarijnS95, line 11: @madsmtm does it appear to you that anything weird is happening? This seems to be [the same implementation as `metal-rs`](https://docs.rs/metal/0.29.0/src/metal/device.rs.html#1513-1520) yet with `objc2` it always returns NULL on an M2 Air. It returns a valid device via `metal-rs`.

MarijnS95: Never mind. Conveniently I just went through a rename in the documentation for https://github.com/madsmtm/objc2/pull/638 and spotted https://docs.rs/objc2-metal/latest/objc2_metal/, we need to set up the linking via the framework exactly.

I expected that to result in a linker/compiler error, not a "returns NULL at runtime" :sweat_smile: 

Should we use that or another PR to hide such functions behind a CFG that turns on linking automatically? Is this behaviour the same for all functions?

MarijnS95: It is also quite funky that `MTLCopyAllDevices()` doesn't suffer from this `CoreAnimation` dependency and can be called directly; probably because it needs CA to figure out which of the devices is the "default" one (though it's easy to just pick the one and only GPU from `MTLCopyAllDevices()` as multi-GPU seems not very common).

madsmtm: > Should we use that or another PR to hide such functions behind a `feature` that turns on linking automatically, as an extra requirement to use `MTLCreateSystemDefaultDevice()` if it's known to always return `NULL` (and be effectively useless) otherwise?

Hmm, I'm not entirely sure. This is mostly just an already documented caveat (see [Apple's official docs](https://developer.apple.com/documentation/metal/1433401-mtlcreatesystemdefaultdevice?language=objc)), and not something handled by Swift either.

The hardest part about this is that `#[link(...)]` attributes are not always passed on to the linker (at least not before https://github.com/rust-lang/rust/issues/121293 is resolved), so users can still run into this problem (e.g. if they compile a static Rust library w. Cargo, and link that with a C project compiled in Xcode that only links to the Metal framework).

MarijnS95: Good point, specifically to note that while `MTLCreateSystemDefaultDevice()` comes from a `link` against the `Metal` framework, and something internal expects us to link to `CoreGraphics` (but `MTLCreateSystemDefaultDevice` isn't linked from `CoreGraphics` itself). I may have implicitly skipped over that.

Jasper-Bekkers (approved):

Jasper-Bekkers:
I've now succesfuly ran with these changes, I think we should go ahead and merge them.

MarijnS95:
https://github.com/Traverse-Research/gpu-allocator/commit/6a2f5214e99403e596cb4dc93cd82f5b2b3f2f8d

> Add 'all' dependency because I can't get it to work without when basing our code on the latest 'objc2-metal' branch

This sounds wrong so I've dropped the commit. If `gpu-allocator` builds on/for Metal correctly, the issue lies elsewhere. Maybe another crate forgot to list the right features? Or there is interop that requires a different set of features?

Let me investigate for a bit before we merge this :pray:

Jasper-Bekkers:
> This sounds wrong so I've dropped the commit. 

Yeah no worries, it's a hackfix for me to be able to use this in my local testing, I can apply it locally. I didn't want to dive into why it was breaking (after having spent some time on it). 

> If gpu-allocator builds on/for Metal correctly, the issue lies elsewhere. Maybe another crate forgot to list the right features? Or there is interop that requires a different set of features?

I'm building on a later commit then what's released, I think the feature setup just changed.

MarijnS95:
> Let me investigate for a bit before we merge this 🙏

Found it: the `git` version that you're patching to hides quite a few of the types used here behindd a new `MTLAllocation` feature since the XCode 16 beta 1 update:

https://github.com/madsmtm/objc2-generated/commit/aadc4edddcf4d59e408ee608a9115bda54d317bf

We can't explicitly add this feature yet because it's not available in the `objc2` dependency here, and `"all"` would kindo fbe papering over it.

---

Let's add `MTLAllocation` in our own codebase (or `all`) and patch it up here when we upgrade to the new version (which I hope counts as a semver break).

Jasper-Bekkers:
Sounds good, lets merge?

MarijnS95:
Ready for merge once CI is green :tada:

Issue #224, opened 2024-04-30
⚠️ NOTICE for `metal` users on `gpu-allocator 0.26` ⚠️ 


MarijnS95:
[`gpu-allocator 0.26`](https://github.com/Traverse-Research/gpu-allocator/releases/tag/0.26.0) was just published with `metal` support! Unfortunately the [`metal-rs` crate that is used](https://github.com/gfx-rs/metal-rs) is heavily unmaintained, and some functions used by `gpu-allocator` are not yet available in the upstream `metal-rs` release. Attempts to contribute have thus far gone unnoticed.

This was worked around by a [`patch` in `Cargo.toml`](https://github.com/Traverse-Research/gpu-allocator/blob/cbf52992b972c6cba6b1ba0530197642614b4154/Cargo.toml#L100-L101) which does not translate to the crates.io release, so anyone turning on the **not-`default`** `metal` feature will get a compilation failure in `gpu-allocator`.

The workaround is to adopt the same `[patch.crates-io]` into your workspace root, while we work to resolve this issue upstream:

https://github.com/Traverse-Research/gpu-allocator/blob/cbf52992b972c6cba6b1ba0530197642614b4154/Cargo.toml#L100-L101

MarijnS95:
A planned upgrade to `objc2` and `objc2-metal` is happening in https://github.com/Traverse-Research/gpu-allocator/pull/225 :tada:

MarijnS95:
This was fixed by using stable `metal 0.29` complete will all necessary changes in the [published `gpu-allocator 0.27` release](https://github.com/Traverse-Research/gpu-allocator/releases/tag/0.27.0).

Jasper-Bekkers:
For any users upgrading that are still on metal-rs, it's relatively trivial to unsafely transmute the structures back and forth as a stopgap solution.

MarijnS95:
Note that the `metal-rs` -> `objc2` conversion will only come in the `gpu-allocator 0.28` release, so you can also stick around on the older version for a little while while still benefiting from the fix for this issue :)

PR #228, opened 2024-05-22
cargo: Replace `metal-rs` patch with `git` dependency

MarijnS95:
`gpu-allocator` is a library crate, meaning that `[patch]`es don't get applied when users depend on our crate, and falsely end up with a broken build while our CI said everything was hunky dory ( https://github.com/Traverse-Research/gpu-allocator/issues/224).  Even though we cannot normally have a direct `git` dependency as that blocks us from publishing to crates.io, the current `gpu-allocator` release (though only for Metal) is broken anyway.  It's relying on functionality that has just now been merged upstream, but still has to make it into a (followup patch) release.

What's worse, the harcoded hash that this was pointing to no longer has a live reference on our fork (maybe due to a force-push), and the CI is now failing to to fetch that commit by hash while build-testing `gpu-allocator`.

Let's bump to a `git` dependency for now, and replace that as soon as we have a workable solution, however that pans out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,8 +77,6 @@ jobs:
           # is also unmaintained for 4 years depends on malloc_buf >=0.0,<0.1.0,
           # allowing the 0.0.6 release to be used (but not the 1.0.0 release).
           cargo update -p malloc_buf --precise 0.0.6
-          # Force 0.1.2, as older versions use the broken `gcc` crate
-          cargo update -p objc_exception --precise 0.1.2
       - name: Cargo clippy with minimal-versions
         run: cargo +stable clippy --workspace --all-targets --features ${{ matrix.features }} --no-default-features -- -D warnings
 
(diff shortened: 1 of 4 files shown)


MarijnS95:
The CI is probably not going to like this change either, because the patch is needed to build `metal`. The upstream PR (https://github.com/gfx-rs/metal-rs/pull/316) isn't merged yet and as explained a `[patch]` or `git` relesae doesn't work anyway, so we might just switch this dep to our custom registry and postpone crates.io releases until `metal-rs` gets all outstanding PRs merged, together with a minor/patch version bump?

Alternatively, could we make the metal code compatible with the latest release (i.e. drop the heap type and acceleration structure allocations) to preempt all this mess?

MarijnS95:
> Alternatively, could we make the metal code compatible with the latest release (i.e. drop the heap type and acceleration structure allocations) to preempt all this mess?

I think we should bite the bullet and do this. Neither d3d12 nor Vulkan provide helpers to turn an `Allocation` into a buffer/texture/accelerationstructure, but give direct access to the underlying heap and offset+size of the region that's provided for this allocation. Besides, it doesn't even publicly expose those basic fields like the other APIs.

Unfortunately once we fix that we're still left with a missing `heap_acceleration_structure_size_and_align_with_size()` to create an `AllocationCreateDesc`.

MarijnS95:
We could also do a quick hack and push a tag to make https://github.com/Traverse-Research/metal-rs/commit/a354c33 valid again, but that *doesn't make any of the other issues outlined here go away*.

tosti007 (approved):

tosti007 (approved):

PR #229, opened 2024-05-24
🦀 Fix Rust 1.78 lints

MarijnS95:
Rust 1.78 is again a bit more strict and better at finding `unused_qualifications` and `dead_code`.  Solve those, and clean up some documentation while at it.

### DIFF
--- a/src/allocator/mod.rs
+++ b/src/allocator/mod.rs
@@ -73,15 +73,7 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug + Sync + Send 
     #[must_use]
     fn supports_general_allocations(&self) -> bool;
     #[must_use]
-    fn size(&self) -> u64;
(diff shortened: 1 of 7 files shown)


src/allocator/mod.rs
```diff
@@ -73,16 +73,8 @@ pub(crate) trait SubAllocator: SubAllocatorBase + std::fmt::Debug + Sync + Send
     #[must_use]
     fn supports_general_allocations(&self) -> bool;
     #[must_use]
-    fn size(&self) -> u64;
```
FlannyH, line 76: Doesn't the visualizer use this function?

MarijnS95: Looks like a silly desync between `d3d12` and `vulkan` where the latter checks the block directly while the former uses the function on the suballocator.

tosti007 (approved):

PR #233, opened 2024-06-14
Increase allowed `windows` version range to `0.53-0.57` and bump MSRV to `1.70`

MarijnS95:


Jasper-Bekkers (approved):

Jasper-Bekkers:
Approving but commits still need to be signed it looks like.

Jasper-Bekkers (approved):

PR #236, opened 2024-07-07
 🪟 Update to windows 0.58 

Jasper-Bekkers:
Depends on #233

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,8 @@ env_logger = "0.10"
 winapi = { version = "0.3.9", features = ["d3d12", "d3d12sdklayers", "dxgi1_6", "winerror", "impl-default", "impl-debug", "winuser", "windowsx", "libloaderapi"] }
 
 [target.'cfg(windows)'.dev-dependencies.windows]
-version = ">=0.53,<=0.58"
+# API-breaks since Windows 0.58 only affect our examples
+version = "0.58"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gpu-allocator"
-version = "0.26.0"
+version = "0.26.4"
```
MarijnS95, line 3: Don't think we should bump it on `main` yet. I've been bumping it on our `traverse-pub` repository (should have used a `-alpha.X` suffix like I did for others) because we're on a "published" `metal-rs` fork. The initial Metal PR for `gpu-allocator` here relied on a PR branch on `metal-rs`, and until they make a release we're stuck with that or a `git` dependency. I can ask if they can make one.

MarijnS95 (approved):
I pushed a commit to require version `0.58` for the examples because it contains two breaking changes. Also dropped the `Cargo.toml` version bump so that we can publish it cleanly later.

Indeed the fork-release should have used a `-` suffix in the version.

Cargo.toml
MarijnS95: A `metal-rs` release has been made and we can get rid of the current situation by merging https://github.com/Traverse-Research/gpu-allocator/pull/240 🎉

MarijnS95 (approved):

PR #239, opened 2024-07-07
Fix `render_memory_block_window`

### LINKED ISSUE #238: Vulkan visualizer's `render_memory_block_window` calls `render_breakdown_ui`
https://github.com/Traverse-Research/gpu-allocator/blob/main/src/vulkan/visualizer.rs#L153

It should probably call `render_memory_block_ui` instead :D

Cool visualizer btw! <3

h3r2tic:
`render_memory_block_window ` was calling `render_breakdown_ui` instead of `render_memory_block_ui`.

Resolves #238

### DIFF
--- a/src/d3d12/visualizer.rs
+++ b/src/d3d12/visualizer.rs
@@ -173,7 +173,7 @@ impl AllocatorVisualizer {
     ) {
         egui::Window::new("Allocator Memory Blocks")
             .open(open)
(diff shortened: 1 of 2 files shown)


Jasper-Bekkers (approved):

Jasper-Bekkers:
Ah - we require signed commits, that's why it's not merging.

PR #241, opened 2024-07-08
❓️ `rename_allocation` via interior mutability

h3r2tic:
Ohai, I'm not sure if this is useful to you, or a desirable change, but I thought I'd share anyway.

In our engine, we currently keep `Allocation`s immutable, so updating their debug names isn't possible without wrapping them in some sorta mutex.

Here, I use `arc-swap` to allow interior mutability for the debug names.

(sorry for the noise in Cargo.toml, it got auto-formatted. I can clean it up if you're interested in the change)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,49 +12,65 @@ keywords = ["vulkan", "memory", "allocator"]
 documentation = "https://docs.rs/gpu-allocator/"
 rust-version = "1.69"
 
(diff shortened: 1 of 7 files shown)


Jasper-Bekkers:
Hey tom I think these changes are useful, if you have time could you rebase them on main & drop the formatting changes in Cargo.toml?

PR #250, opened 2024-10-03
✂️ metal: Qualified use cleanup

MarijnS95:
Some items are imported in scope with `as _` and then again referenced using fully-qualified syntax.  Clean that up, and drop the `use objc2_metal as metal` rename while at it.

Turns out Mac's aren't good at pushing code after all because this was scheduled to be part of the previous PR, but somehow missed out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,6 @@ allocator.free(allocation).unwrap();
 
 ```rust
 use gpu_allocator::metal::*;
-use objc2_metal as metal;
 let mut allocator = Allocator::new(&AllocatorCreateDesc {
     device: device.clone(),
     debug_settings: Default::default(),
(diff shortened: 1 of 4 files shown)


MarijnS95:
> and drop the `use objc2_metal as metal` rename while at it.

Could do the same in the example and docs probably?

Jasper-Bekkers:
> > and drop the `use objc2_metal as metal` rename while at it.
> 
> Could do the same in the example and docs probably?

Would be nice, all of the structs in objc2_metal start with MTL anyway to disambiguate them.

Jasper-Bekkers (approved):

MarijnS95:
> start with MTL anyway to disambiguate them.

I was thinking of that more as a reason for importing them in scope rather than renaming the (optional) qualification. The `as metal` rename would make it easier to convert away from the `metal-rs` crate, but there are so many renames that it's not very helpful anyway :)

PR #251, opened 2024-10-14
🦋 metal: Remove inconsistent descriptor creators from `Allocation`

Jasper-Bekkers:
These helpers don't really belong in the metal implementation of GPU allocator, instead, they belong in the client APIs using gpu-allocator.

This issue surfaced in our codebase because the `make_acceleration_structure` function didn't pass through all the required information.

### DIFF
--- a/src/metal/mod.rs
+++ b/src/metal/mod.rs
@@ -39,55 +39,24 @@ pub struct Allocation {
 }
 
 impl Allocation {
-    pub fn heap(&self) -> &ProtocolObject<dyn MTLHeap> {
+    /// Return the backing heap for this allocation
+    pub unsafe fn heap(&self) -> &ProtocolObject<dyn MTLHeap> {
         &self.heap
     }
 
-    pub fn make_buffer(&self) -> Option<Retained<ProtocolObject<dyn MTLBuffer>>> {
-        let resource = unsafe {
-            self.heap.newBufferWithLength_options_offset(
-                self.size as usize,
-                self.heap.resourceOptions(),
-                self.offset as usize,
-            )
-        };
-        if let Some(resource) = &resource {
-            if let Some(name) = &self.name {
-                resource.setLabel(Some(&NSString::from_str(name)));
-            }
-        }
-        resource
+    /// Returns the size of the allocation
+    pub fn size(&self) -> u64 {
+        self.size
     }
 
-    pub fn make_texture(
-        &self,
-        desc: &MTLTextureDescriptor,
-    ) -> Option<Retained<ProtocolObject<dyn MTLTexture>>> {
-        let resource = unsafe {
-            self.heap
-                .newTextureWithDescriptor_offset(desc, self.offset as usize)
-        };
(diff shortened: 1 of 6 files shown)


MarijnS95:
Glad to see this finally happening after proposing it months ago.  These things shouldn't have been there as they're inconsistent with our Vulkan and Dx12 backend, and above all provoked a terrible `.allocate().make_xxx()` pattern where the returned `Allocation` is never stored and hence never able to be `.free()`d anymore, causing massive memory leaks.

I've tried to retitle this PR to describe the specific bad pattern that you're removing, happy to see this reclarified/rewritten further if you have better ideas :+1:

manon-traverse (approved):

src/metal/mod.rs
```diff
@@ -39,55 +39,24 @@ pub struct Allocation {
 }
 
 impl Allocation {
-    pub fn heap(&self) -> &ProtocolObject<dyn MTLHeap> {
+    /// Return the backing heap for this allocation
+    pub unsafe fn heap(&self) -> &ProtocolObject<dyn MTLHeap> {
```
MarijnS95, line 43: Do we need a `/// # Safety` to explain this one?

src/metal/mod.rs
```diff
@@ -39,55 +39,24 @@ pub struct Allocation {
...
-            self.heap
-                .newTextureWithDescriptor_offset(desc, self.offset as usize)
-        };
-        if let Some(resource) = &resource {
-            if let Some(name) = &self.name {
-                resource.setLabel(Some(&NSString::from_str(name)));
-            }
-        }
-        resource
+    /// Returns the offset of the allocation on the [`ID3D12Heap`].
```
MarijnS95, line 52:
```suggestion
    /// Returns the offset of the allocation on the [`MTLHeap`].
```

src/metal/mod.rs
```diff
@@ -39,55 +39,24 @@ pub struct Allocation {
...
+    /// When creating a placed resources, this offset needs to be supplied as well.
```
MarijnS95, line 53:
```suggestion
    /// When creating a placed resource, this offset needs to be supplied as well.
```

src/metal/mod.rs
```diff
@@ -39,55 +39,24 @@ pub struct Allocation {
...
-                .newAccelerationStructureWithSize_offset(self.size as usize, self.offset as usize)
-        };
-        if let Some(resource) = &resource {
-            if let Some(name) = &self.name {
-                resource.setLabel(Some(&NSString::from_str(name)));
-            }
-        }
-        resource
+    pub fn name(&self) -> Option<&str> {
+        self.name.as_ref().map(|s| s.as_ref())
```
MarijnS95, line 59: Not sure, but IIRC:
```suggestion
        self.name.as_deref()
```

src/metal/mod.rs
MarijnS95: Since this one was copied from `d3d12/mod.rs`, mind fixing the typo there as well?

PR #252, opened 2024-10-30
wrap `Arc<Backtrace>` in Option, saving memory when backtraces are disabled

Firestar99:
This PR wraps all `Arc<Backtrace>` in Option, as to not have to allocate an Arc of a disabled Backtrace each time GPU memory is allocated with backtraces are disabled. This is probably a premature optimization, but it was so easy to implement I couldn't resist :D

### DIFF
--- a/src/allocator/free_list_allocator/mod.rs
+++ b/src/allocator/free_list_allocator/mod.rs
@@ -32,7 +32,7 @@ pub(crate) struct MemoryChunk {
     pub(crate) allocation_type: AllocationType,
     pub(crate) name: Option<String>,
     /// Only used if [`crate::AllocatorDebugSettings::store_stack_traces`] is [`true`]
-    pub(crate) backtrace: Arc<Backtrace>,
+    pub(crate) backtrace: Option<Arc<Backtrace>>,
     next: Option<std::num::NonZeroU64>,
     prev: Option<std::num::NonZeroU64>,
 }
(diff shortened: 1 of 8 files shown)


manon-traverse (approved):
Looks like a nice small optimization. Thanks for your contribution! :D

MarijnS95:
I recall checking **long ago** that `Backtrace::Captured { .. }` used to be quite small back in the day, is that no longer the case?

Regardless, this saves us some bytes in the `Arc` too so I guess it's nice to have. There's nothing to do about the ambiguity between `None` and `Some(Arc::new(Backtrace::Disabled))` but it is what it is :)

src/allocator/free_list_allocator/mod.rs
```diff
@@ -384,7 +384,10 @@ impl SubAllocator for FreeListAllocator {
                 chunk.offset,
                 chunk.allocation_type,
                 name,
-                chunk.backtrace
+                chunk
+                    .backtrace
+                    .as_ref()
+                    .map_or(&Backtrace::disabled(), |b| b)
```
MarijnS95, line 390: Aren't these equivalent to:
```suggestion
                chunk.backtrace.as_deref().unwrap_or(&Backtrace::disabled())
```

MarijnS95 (commented):
Looking at this PR I still wonder why the `Backtrace` is wrapped in an `Arc`; in e45e25afdce4d747db5c6d93f321005808da42b8 I wrote that that is to support "`Clone`ing" the `Backtrace`, but glancing at the implementation that is only because of mistakes in the ownership transfer in our `fn allocate()` code and copies for `AllocationReport` which should either not receive a `Backtrace` at all, or borrow it temporarily while the caller can _view_ the backtrace.

src/allocator/free_list_allocator/mod.rs
Firestar99: implemented in both allocators

PR #256, opened 2024-12-10
Remove stale and UB `winapi` compatibility layer

### LINKED ISSUE #255: Maybe Unsound in from_winapi_d3d12_resource_desc
Hello, thank you for your contribution in this project, I am scanning the unsoundness problem in rust project.
I notice the following code:
```rust
pub fn from_winapi_d3d12_resource_desc(
        device: *const winapi_d3d12::ID3D12Device,
        desc: &winapi_d3d12::D3D12_RESOURCE_DESC,
        name: &'a str,
        location: MemoryLocation,
    ) -> Self {
        let device = device.as_windows();
        // Raw structs are binary-compatible
        let desc = unsafe {
            std::mem::transmute::<&winapi_d3d12::D3D12_RESOURCE_DESC, &D3D12_RESOURCE_DESC>(desc)
        };
        let allocation_info =
            unsafe { device.GetResourceAllocationInfo(0, std::slice::from_ref(desc)) };
        let resource_category: ResourceCategory = desc.into();

        AllocationCreateDesc {
            name,
            location,
            size: allocation_info.SizeInBytes,
            alignment: allocation_info.Alignment,
            resource_category,
        }
    }
```
Considering that `pub mod d3d12`, and `from_winapi_d3d12_resource_desc`  is also a pub function. I assume that users can directly call this function. This potential situation could result in `device.GetResourceAllocationInfo(0, std::slice::from_ref(desc))` operating on a null pointer, I am not sure what will `null_pointer.GetResourceAllocationInfo` do but I guess it might trigger undefined behavior (UB). For safety reasons, I felt it necessary to report this issue. If you have performed checks elsewhere that ensure this is safe, please don’t take offense at my raising this issue.
I suggest Several possible fixes: 
1. If there is no external usage for `from_winapi_d3d12_resource_desc` it should not marked as `pub`. 
2. `from_winapi_d3d12_resource_desc` method should add additional check for null pointer.
3. mark `from_winapi_d3d12_resource_desc` method as unsafe and proper doc to let users know that they should provide valid Pointers.

MarijnS95:
Fixes #255

With the `windows` migration we introduced a `winapi` compatibility layer to make it easier to use `winapi` types with `gpu-allocator`.  As most of the ecosystem has switched to `windows` it is now time to sunset these helpers which are otherwise tedious to test and maintain, not to mention contain long(er) standing unsoundness bugs.


Jasper-Bekkers (approved):

MarijnS95:
Yikes this new presser code is causing some new clippy lints... and as far as I remember didn't even resolve our original soundness issues :sweat_smile:

Jasper-Bekkers (approved):

MarijnS95:
@Jasper-Bekkers in order to merge this, we must remove the branch-protection rules that expect a check with `public-winapi` in the title to succeed, because that feature no longer exists :+1:

PR #257, opened 2024-12-23
add `Allocation::name` getter

Firestar99:
What the title says, nothing special. I don't want to store yet another copy of the same string, it would be nice if I could just query it instead.

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -258,6 +258,10 @@ impl Allocation {
     pub fn is_null(&self) -> bool {
         self.chunk_id.is_none()
     }
+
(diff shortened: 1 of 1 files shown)


MarijnS95:
Should the PR title be prefixed with `vulkan:` or are you intending to add this to the other backends (dx12 and Metal) as well?

Jasper-Bekkers:
Looking at this, does anyone know why it's `name: Option<Box<str>>` and not `name: Option<String>`?

MarijnS95:
@Jasper-Bekkers that is described in https://github.com/Traverse-Research/gpu-allocator/pull/70 :)

PR #258, opened 2025-01-08
cargo: Bump `windows` crate range to `0.53-0.59`

MarijnS95:
https://github.com/microsoft/windows-rs/releases/tag/0.61.0 (tag is strangely on a much higher version...)

Note that `windows 0.59` bumps the MSRV to `1.74`, but we still support this crate dependency all the way back to `0.53`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -44,7 +44,6 @@ rustflags = [
     "-Wclippy::match_wild_err_arm",
     "-Wclippy::match_wildcard_for_single_variants",
     "-Wclippy::mem_forget",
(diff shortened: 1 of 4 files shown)


Jasper-Bekkers (approved):

ErichDonGubler:
Seems like a couple of things are coming up in CI:

1. A new Clippy warning is firing for a lifetime that can be elided, looks easy.
2. Looks like the MSRV checks need to have their version updated. I think the MSRV for 0.59 is 1.74, but the tests might also be persuaded to pass by resolving `windows` to a lower version.

Issue #159, opened 2023-03-23
Allocator type is quite large


i509VCB:
The allocator type is quite large. This is due to holding a `ash::Device`.

Summary of -Z print-type-sizes (`cargo +nightly rustc -- -Z print-type-sizes`):
```
print-type-size type: `gpu_allocator::vulkan::Allocator`: 1552 bytes, alignment: 8 bytes
print-type-size     field `.device`: 1488 bytes
print-type-size     field `.buffer_image_granularity`: 8 bytes
print-type-size     field `.memory_types`: 24 bytes
print-type-size     field `.memory_heaps`: 24 bytes
print-type-size     field `.debug_settings`: 6 bytes
print-type-size     end padding: 2 bytes
```

MarijnS95:
Happy to accommodate and wrap this in an `Arc` as suggested in https://github.com/ash-rs/ash/issues/731#issuecomment-1493382084 - we do that internally in our codebase as well and have a `device: device.as_ref().clone()` with comment on `AllocatorCreateDesc` to duplicate the function pointers. Let's settle https://github.com/ash-rs/ash/issues/731 before definitively doing so though.

speedy-lex:
I've made a PR which uses references (#259)

PR #259, opened 2025-01-08
Store Vulkan device as a reference to avoid huge struct sizes

speedy-lex:
Unfortunately this change is not backwards compatible.

### DIFF
--- a/examples/vulkan-buffer.rs
+++ b/examples/vulkan-buffer.rs
@@ -85,7 +85,7 @@ fn main() {
     // Setting up the allocator
     let mut allocator = Allocator::new(&AllocatorCreateDesc {
         instance: instance.clone(),
-        device: device.clone(),
+        device: &device,
         physical_device: pdevice,
         debug_settings: Default::default(),
         buffer_device_address: false,
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -52,9 +52,9 @@ unsafe impl Send for SendSyncPtr {}
 // `mapped_ptr` themselves.
 unsafe impl Sync for SendSyncPtr {}
 
-pub struct AllocatorCreateDesc {
+pub struct AllocatorCreateDesc<'a> {
     pub instance: ash::Instance,
-    pub device: ash::Device,
+    pub device: &'a ash::Device,
     pub physical_device: vk::PhysicalDevice,
     pub debug_settings: AllocatorDebugSettings,
     pub buffer_device_address: bool,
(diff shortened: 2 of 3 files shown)


Jasper-Bekkers:
Could you provide some context about why this change is required, and required to break backwards compatibility (for example: can't we just put the device in a Box and not have to have the lifetime on Allocator)?

speedy-lex:
This change reduces the `Allocator` type size from 1584 bytes to 104 bytes as you've probably seen.
>why this change is required
>
It's not required, just a solution to #159.
>can't we just put the device in a Box
>
Wrapping `ash::Device` in a `Box` still has the 1.4Kib of pointers duplicated, just that it would be on the heap.
>and required to break backwards compatibility
>
Unfortunately I'm not sure if we can just wrap it in an `Arc` to avoid lifetimes, as https://github.com/ash-rs/ash/issues/731 hasn't been resolved definitively.
Using a reference is good since it still lets the user use other libraries smoothly (as opposed to managing both an `Arc` and an owned `ash::Device` for other libraries). 
Drawbacks include:
- No backwards compatibility. We can't do anything about this without duplicating the device since `Allocator::new` takes `AllocatorCreateDesc` which _owns_ an `ash::Device`. The only other option is to leave a device in `AllocatorCreateDesc` and make it outlive the `Allocator`, but this doesn't seem like desirable behaviour.
- Lifetimes being added to `Allocator` and `AllocatorCreateDesc`. This shouldn't be an issue though since your main `ash::Device` (and more specifically the vulkan object) has to outlive the allocator anyways.

I think references are a good middle-ground between wrapping an `ash::Device` and sharing it (impeding on other libraries) or huge struct sizes from owning it (references will be smaller and _faster_ than `Arc`'s or `Box`es).

Feel free to merge when https://github.com/ash-rs/ash/issues/731 is resolved or even now

MarijnS95:
A reference won't work in a self-referential struct. How would we denote the lifetime in this instance?

```rust
struct Device {
    device: ash::Device,
    allocator: gpu_allocator::vulkan::Allocator<'self_device>,
}
```

Here `'self_device` is the lifetime of `Device::device`.

This is a very real scenario for implementations and abstractions around Vulkan.

speedy-lex:
Hmm, I see. It's either `Arc`'s or owned :/

MarijnS95:
If we really want to, we could use a `Cow<>`. That should allow the user to set a `'static` lifetime when they provided a `Cow::Owned(device)` (but then again doesn't allow a `Box` or `Arc`) :sweat_smile:

speedy-lex:
Nothing works here.
A real solution would be something like `Drop` trait on `ash::Device` but then allocation callbacks can't work without being passed in earlier (`Drop` lets `Arc` work normally). If I think about it, I'm not even sure everyone would be happy with a `Drop` implementation.

MarijnS95:
I don't think that "nothing works" here, just that whichever way we pick is going to be opinionated unless we allow the user to store any form of `ash::Device`, perhaps:

```rust
enum AshDevice<'a> {
    Arc(Arc<ash::Device>),
    Box(Box<ash::Device>),
    // The below two effectively form a Cow:
    Owned(ash::Device),
    Borrowed(&'a ash::Device),
}
```

But the caller still has to ensure they free all their allocations and drop their `Allocator` before freeing the device (and likewise consider destruction order of many other Vulkan objects).

We could also decide to store just `ash::vk::Device` and the few function pointers that we actually need, but will then have to call the raw functions ourselves.

---

Neither do I think that `ash` should implement ownership / `Drop` semantics. Regardless, whether `ash` stores these expensive function-pointer tables on the heap, or within the `struct` is orthogonal to implementing ownership management via `Drop`. With this large, owned struct, the user has maximum flexibility to decide how they want to store their Vulkan device and structure their code. They could for example wrap their entire `struct Device` (i.e. the one I suggested above) in an `Arc` and not pay the "cost" for a second `Arc` inside `ash`.

At the same time that design doesn't work at all with the current design of `Allocator`, as you found out and what you made this PR for - it requires a full clone of all function pointers.

---

Given your comment above that this is "just a solution to #159", does that mean you are not impacted by this but only wanted to address this because of seeing a "misleadingly trivial" open issue? I think we're all totally open to change our design and approach here to accommodate most usage patterns in an efficient way (also why #260 was opened), but need real-world usage examples to decide which of these many solutions is the right way forward. Thanks for bringing this to our attention once again!

PR #261, opened 2025-01-24
🍳 Upgrade to `objc2 0.6` and `objc2-metal 0.3`

MarijnS95:
The `MTLCreateSystemDefaultDevice()` function finally has a high-level wrapper that constructs a `Retained` protocol object for us.

Note also that a new `MTLAllocation` interface was added to Metal.  Even if that interface is _optional_, current `objc2` design requires this trait to be part of the hierarchy (and the feature explicitly enabled) to get access to all "descendant" interfaces (`MTLResource`, which provides `MTLBuffer` and `MTLTexture`). There are upstream plans to see if this can be simplified: https://github.com/madsmtm/objc2/issues/663.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ name: CI
 
 jobs:
   check_msrv:
-    name: Check MSRV (1.70.0)
+    name: Check MSRV (1.71.0)
     strategy:
       matrix:
         include:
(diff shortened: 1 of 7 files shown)


Jasper-Bekkers (approved):

Jasper-Bekkers (approved):

MarijnS95:
Oh - the MSRV check has been renamed to `1.71` to match the bump... and branch protection rules need to be updated. Should we remove the version from the title before we do that?

Jasper-Bekkers:
Rebase went fine but it looks like commits aren't signed.

MarijnS95:
Hmm seems GitHub no longer signs commits with its own key on rebase (for the better, probably).

PR #264, opened 2025-01-31
👩‍🏭Update to objc2-metal 0.3 (requires MSRV bump to 1.71)

Jasper-Bekkers:
We need this as a dependency going forward instead; and this requires some minor changes.

https://docs.rs/objc2-metal/latest/objc2_metal/trait.MTLHeap.html

Notice: this requires an MSRV bump for objc2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ name: CI
 
 jobs:
   check_msrv:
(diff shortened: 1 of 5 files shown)


MarijnS95:
Dupe of https://github.com/Traverse-Research/gpu-allocator/pull/261 I guess. If you update the branch protection rules it can be merged.

PR #265, opened 2025-02-12
Add `no_std` support

CrazyboyQCD:


MarijnS95:
Almost forgot to mention: can you please drop the `chore:` prefix from your commit messages? It's not used by this project and has no meaning.

> a routine task, especially a household one.

I don't think a one-off `no_std` conversion in a crate is considered a "routine task" :slightly_smiling_face:

Cargo.toml
```diff
@@ -12,26 +12,24 @@ keywords = ["vulkan", "memory", "allocator"]
...
 
 [package.metadata.docs.rs]
 all-features = true
 
 [dependencies]
-log = "0.4"
-thiserror = "1.0"
-presser = { version = "0.3" }
+log = { version = "0.4", default-feature = false }
+thiserror = { version = "2.0", default-feature = false }
```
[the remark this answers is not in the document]
CrazyboyQCD: `std`  can be disabled in `thiserror`  since v2, but its msrv is 1.81+ https://github.com/dtolnay/thiserror/pull/373.

But we can use [derive_more](https://github.com/JelteF/derive_more) instead, its msrv is 1.75+.

CrazyboyQCD:
`msrv` is required to bump to either 1.75+(if we switch from `thiserror` to `derive_more`) or 1.81+ to get CI work.

src/allocator/dedicated_block_allocator/mod.rs
```diff
@@ -107,7 +118,23 @@ impl SubAllocator for DedicatedBlockAllocator {
             self.size,
             name,
             self.backtrace
-        )
+        );
+        #[cfg(not(feature = "std"))]
+        log!(
```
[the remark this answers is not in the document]
CrazyboyQCD: We can't map backtrace here since we remove the field and import by feature, what I do is to format it ahead by feature.

src/allocator/free_list_allocator/mod.rs
```diff
@@ -1,16 +1,25 @@
 #![deny(unsafe_code, clippy::unwrap_used)]
 
-#[cfg(feature = "visualizer")]
-pub(crate) mod visualizer;
+use alloc::{
+    borrow::ToOwned,
+    string::{String, ToString},
+    vec::Vec,
+};
```
MarijnS95, line 8:
```suggestion
```

src/lib.rs
```diff
@@ -212,6 +212,21 @@
...
+    clippy::std_instead_of_alloc,
+    clippy::std_instead_of_core
+)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[macro_use]
+extern crate alloc;
+
+#[cfg(all(feature = "std", feature = "hashbrown"))]
+compile_error!("\"hashbrown\" feature should not be enabled in \"std\" environment.");
```
MarijnS95, line 226: Just curious if you've seen the second paragraph of https://github.com/Traverse-Research/gpu-allocator/pull/265#discussion_r1952226867, and have made a cognitive choice whether it's possible/sensible to support users "accidentally" enabling both features.

To repeat, if I have a library crate with `gpu-allocator = { default-features = false, features = ["hashbrown"] }`, and a user later depends on my crate but also wants to use `gpu-allocator` **with the `"std"` feature**, they'll get an error when instead we might want to keep using `hashbrown` and/or only fall back when `"std"` is enabled but `"hashbrown"` is not, like this?

```rust
#[cfg(all(feature = "std", not(feature = "hashbrown")))]
use std::collections::{HashMap, HashSet};
```

MarijnS95 (commented):
Nice update!

Only two things remaining: the formatting of the reports is broken, please test your changes to these before pushing them. I've suggested how to fix this in one file.

The previous review contained a request to update the CI to ensure that we test both with and without these new feature combinations, that would be very helpful to prevent breakage going forward. Right now we always build with specific backends and `--no-default-features`, meaning neither `"hashbrown"` nor `"std"` (nor a combination) ever gets exercised properly.

Cargo.toml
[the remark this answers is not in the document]
CrazyboyQCD: @MarijnS95 
Should we use [derive_more](https://github.com/JelteF/derive_more) for bumping lower MSRV?

.github/workflows/ci.yml
```diff
@@ -9,11 +9,11 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            features: vulkan
+            features: std,vulkan
```
MarijnS95, line 12: It looks like you added `std` to each matrix entry, which means our CI never tests **without `std` support**??

Isn't that exactly what you want, so that `no_std` support keeps being tested and working going forward?

MarijnS95: More specifically, I think we should see a few extra combinations:

(A backend like `vulkan`, plus:)

1. No default features, plus;
2. `hashbrown` feature;
3. `std` feature;
4. `hashbrown` + `std` feature.

That way all nontrivial combinations are tested.

Cargo.toml
```diff
@@ -98,11 +99,13 @@ name = "metal-buffer"
...
 
 [features]
+std = ["presser/std"]
 visualizer = ["dep:egui", "dep:egui_extras"]
 vulkan = ["dep:ash"]
 d3d12 = ["dep:windows"]
 metal = ["dep:objc2", "dep:objc2-metal", "dep:objc2-foundation"]
 # Expose helper functionality for winapi types to interface with gpu-allocator, which is primarily windows-rs driven
 public-winapi = ["dep:winapi"]
+hashbrown = ["dep:hashbrown"]
```
MarijnS95, line 109: Let's document what this does:
```suggestion
# Enables the FreeListAllocator when `std` is not enabled by using the `hashbrown` crate
hashbrown = ["dep:hashbrown"]
```

MarijnS95 (dismissed):
Almost there :crossed_fingers:. We just need proper and complete CI to ensure that the changes are correct as written, and most importantly: won't regress `no_std` support and feature combinations in the future.

.github/workflows/ci.yml
```diff
@@ -9,11 +9,23 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            features: std,vulkan
+            features: vulkan,hashbrown
```
MarijnS95, line 12: So from this I gauge that `--no-default-features -F vulkan` in isolation doesn't work (or any other backend for that matter) because they all unconditionally require `FreeListAllocator`?

In that case it might make sense to drop the `#[cfg(any(feature = "std", feature = "hashbrown"))]` in `allocator/mod.rs` and add a ``compile_error!("Either `std` or `hashbrown` must be enabled");`` to match. What do you think?

I'm not exactly sure what's best since this crate funnily _does_ compile with `--no-default-features` (meaning no backends, but only a few types in `lib.rs`... unsure if this is useful for certain library crates).

CrazyboyQCD: It was this snippet that `vulkan` can't work alone.
https://github.com/Traverse-Research/gpu-allocator/blob/3e453edf7478fd39418f5298997cda17d74d6d03/src/vulkan/mod.rs#L418-L425

MarijnS95: Correct, that's what failed when testing this PR locally as well, so it's good to have a clear `compile_error!()` with instructions for the user instead of letting them run into random compiler errors :+1:

src/lib.rs
```diff
@@ -212,6 +212,21 @@
...
+    clippy::std_instead_of_alloc,
+    clippy::std_instead_of_core
+)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[macro_use]
+extern crate alloc;
+
+#[cfg(all(not(feature = "std"), feature = "visualizer"))]
+compile_error!("Cannot enable \"visualizer\" feature in \"no_std\" environment.");
```
MarijnS95, line 226:
```suggestion
compile_error!("Cannot enable `visualizer` feature in `no_std` environment.");
```

Let's match these to the below.

MarijnS95 (approved):

Cargo.toml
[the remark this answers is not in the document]
MarijnS95: Hmm we still need to look into this. IIRC there was some pushback on bumping to `thiserror` in some of our other repos because the crate is used incredibly frequently and requires an update throughout many (of our) crates in the ecosystem.

OTOH I don't think we require them to match to compile our code, it's just a duplicate dependency we can take on in `cargo-deny`.

.github/workflows/ci.yml
MarijnS95: Thinking out loud (sorry for all these consecutive review comments), manually listing out all the features kind of defeats the purpose of having a `matrix`, which is defined to generate a "Cartesian product" between all possible combinations.

We initially used specific combinations via `include` to ensure only the right `os`+`feature` set lined up. Now that we have more feature combinations, I think we should split those out to their own "object" in the matrix. I've pushed a sample of that to this PR to try it out and see how that would work, ensuring all combinations are adequately tested while making it easier to read.

MarijnS95: Finally, and this requies some thinking on my end, maybe we don't want to test _all combinations_ for the MSRV test _and_ the `clippy` checks _and_ the `cargo test`s, that's a bit overzealous and spawns a massive amount of jobs that all take CI time... Maybe we can recombine some and reuse intermediate artifacts as well.

CrazyboyQCD: @MarijnS95 
Sorry I accidently overried your commit whien fixed the `std` usages, I don't know how to recover it.

MarijnS95 (commented):
Nice! I trust you'll be able to follow up on the remaining CI errors so that we can merge this! Do you think support for `no_std` needs a nice shout-out (including how to use it with `default-features = false` and that `"hashbrown"` must be used?) in the `README`?

I'll meanwhile look into how far we are with a `thiserror 2.0` migration in our dependency tree, that might be more useful than to switch crates.

MarijnS95 (approved):

README.md
```diff
@@ -167,9 +168,39 @@ drop(resource);
...
+[dependencies]
+gpu-allocator = { version = "0.27", default-features = false }
+```
+
 ## Minimum Supported Rust Version
 
-The MSRV for this crate and the `vulkan`, `d3d12` and `metal` features is Rust 1.71.  Any other features such as the `visualizer` (with all the `egui` dependencies) may have a higher requirement and are not tested in our CI.
+The MSRV for this crate and the `vulkan`, `d3d12` and `metal` features is Rust **1.71**.
+
+The `no_std` support requires version above **1.81** beacuase `no_std` support of dependency `thiserror` requires `core::error::Error` which is stabalized in **1.81**.
```
MarijnS95, line 201:
```suggestion
The `no_std` support requires Rust **1.81** or higher because `no_std` support of dependency `thiserror` requires `core::error::Error` which is stabilized in **1.81**.
```

src/visualizer/memory_chunks.rs
```diff
@@ -80,7 +80,6 @@ pub(crate) fn render_memory_chunks_ui<'a>(
                         if cursor_idx < data.len() {
                             bytes_required = data[cursor_idx].size;
                         }
-                        continue;
```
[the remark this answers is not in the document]
MarijnS95: Oh never mind, it's not a `break`, and it's a new clippy lint that points this out: https://github.com/Traverse-Research/gpu-allocator/actions/runs/14488776237/job/40640167629?pr=269

release.toml
```diff
@@ -7,5 +7,7 @@ publish = false
 
 pre-release-replacements = [
   { file = "README.md", search = "gpu-allocator = .*", replace = "{{crate_name}} = \"{{version}}\"" },
+  { file = "README.md", search = "gpu-allocator = { version = .*,", replace = "{{crate_name}} = { version = \"{{version}}\"," },
   { file = "README.tpl", search = "gpu-allocator = .*", replace = "{{crate_name}} = \"{{version}}\"" },
+  { file = "README.tpl", search = "gpu-allocator = { version = .*,", replace = "{{crate_name}} = { version = \"{{version}}\"," },
```
MarijnS95, line 12: Now that I look at this, the original rules will likely replace your `gpu-allocator = { version = "0.27", default-features = false, features = ["hashbrown", "other features"] }` with a `gpu-allocator = "0.xx"`...

Perhaps we should limit (and test, please) those to `search = "gpu-allocator = \".*\""`?

CrazyboyQCD: I changed the pattern to match `gpu-allocator = { version = "version"`.

MarijnS95: After testing this, that pattern doesn't work because `{` is the start of a repetition quantifier which must be explicitly escaped. Secondly, the `".*"` is greedy which would remove the remainder of the string in the `version = "...", default-features = ["...", "..."]`.

As requested, I've also reverted the change that makes everything a `{ version = "..." }` which is quite easy to parse separately. Now `cargo release` generates the correct diff as expected.

MarijnS95:
I've squashed everything down into a single commit because our branch protection requires every individual commit to have been signed. Even if squash-merging will squash and resign the entire PR with GitHub's "web workflow" signature **anyway** 🤷 

Since I contributed to and approve the final result, it makes sense for me to sign such a final result :)

Jasper-Bekkers:
Should we go ahead and merge this?

src/lib.rs
```diff
@@ -254,12 +264,14 @@ pub struct AllocatorDebugSettings {
...
     /// or other memory allocations, but storing stack traces has a RAM overhead so should be disabled
     /// in shipping applications.
+    #[cfg(feature = "std")]
     pub store_stack_traces: bool,
     /// Log out every allocation as it's being made with log level Debug, rather spammy so off by default
     pub log_allocations: bool,
     /// Log out every free that is being called with log level Debug, rather spammy so off by default
     pub log_frees: bool,
     /// Log out stack traces when either `log_allocations` or `log_frees` is enabled.
+    #[cfg(feature = "std")]
```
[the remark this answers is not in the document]
MarijnS95: As it turns out this does not work as expected outside of the crate, because of https://rust-lang.github.io/rfcs/2008-non-exhaustive.html#functional-record-updates: it assumes that a private field might be added which cannot be set with field-record-update even if only public fields are set and the rest is provided with spread syntax (`..Default::default()`). In other words, the only valid syntax now seems to be:

```rust
let mut settings = AllocatorDebugSettings::default();
settings.log_memory_information = true;
```

Let me think about possible solutions and alternatives, before likely removing `#[non_exhaustive]` again.

PR #273, opened 2025-06-12
metal: Create a global residency set, holding all allocated heaps

MarijnS95:
`gpu-allocator` creates heaps on which callers can allocate ranges and create "placed" resources like textures, buffers and acceleration structures.  These individual resources, or the heaps as a whole, need to be made resident on the command buffer or even globally on an entire queue.

In the previous API those heaps had to be made resident on individual command *encoders* with `useHeap(s):` (making an entire heap resident perfectly matches a bindless design, as opposed to making every individual resource -either placed on the heap or allocated separately-
resident with `useResource(s):`).  Worse, this API only applies `MTLResourceUsageRead` (exluding `RenderTarget` and `ShaderWrite` textures) which would disallow any resources on the heap to be written.

Now with `MTLResidencySet` multiple heaps can be made resident with one call, defeating the performance overhead of individually "using" all heaps on *every* command *encoder*.  But without tracking this inside `gpu-allocator`, users of our crate still have to manually rebuild this `MTLResidencySet` each time they change their allocations, without knowing when `gpu-allocator` created or destroyed a heap.

By managing a single updated `MTLResidencySet` in `gpu-allocator`, callers can simply call `.commit()` on this object right before they submit command buffers referencing resources on these heaps, as long as they have the residency set attached to the queue in question or "used" on the command buffer that is being submitted.  This removes all the performance overhead of repeatedly creating `MTLResidencySet`s, which otherwise defeats the purpose of it over plain `useHeap(s):` call(s).

### DIFF
--- a/src/d3d12/mod.rs
+++ b/src/d3d12/mod.rs
@@ -475,28 +475,22 @@ impl MemoryType {
 
         mem_block.sub_allocator.free(allocation.chunk_id)?;
 
-        if mem_block.sub_allocator.is_empty() {
-            if mem_block.sub_allocator.supports_general_allocations() {
-                if self.active_general_blocks > 1 {
-                    let block = self.memory_blocks[block_idx].take();
-                    if block.is_none() {
-                        return Err(AllocationError::Internal(
-                            "Memory block must be Some.".into(),
-                        ));
-                    }
-                    // Note that `block` will be destroyed on `drop` here
+        // We only want to destroy this now-empty block if it is either a dedicated/personal
+        // allocation, or a block supporting sub-allocations that is not the last one (ensuring
+        // there's always at least one block/allocator readily available).
+        let is_dedicated_or_not_last_general_block =
+            !mem_block.sub_allocator.supports_general_allocations()
+                || self.active_general_blocks > 1;
+        if mem_block.sub_allocator.is_empty() && is_dedicated_or_not_last_general_block {
+            let block = self.memory_blocks[block_idx]
+                .take()
+                .ok_or_else(|| AllocationError::Internal("Memory block must be Some.".into()))?;
 
-                    self.active_general_blocks -= 1;
-                }
-            } else {
-                let block = self.memory_blocks[block_idx].take();
-                if block.is_none() {
-                    return Err(AllocationError::Internal(
-                        "Memory block must be Some.".into(),
-                    ));
-                }
-                // Note that `block` will be destroyed on `drop` here
+            if block.sub_allocator.supports_general_allocations() {
+                self.active_general_blocks -= 1;
             }
+
+            // Note that `block` will be destroyed on `drop` here
         }
 
         Ok(())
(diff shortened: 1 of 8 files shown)


src/metal/mod.rs
```diff
@@ -373,28 +388,25 @@ impl MemoryType {
...
-                if self.active_general_blocks > 1 {
-                    let block = self.memory_blocks[block_idx].take();
-                    if block.is_none() {
-                        return Err(AllocationError::Internal(
-                            "Memory block must be Some.".into(),
-                        ));
-                    }
-                    // Note that `block` will be destroyed on `drop` here
+        if mem_block.sub_allocator.is_empty()
+            && (!mem_block.sub_allocator.supports_general_allocations()
```
Jasper-Bekkers, line 392: This complex conditional is significantly less readable then what was there before with the multiple if statements. Please revert.

src/metal/mod.rs
```diff
@@ -427,10 +439,19 @@ impl Allocator {
             }),
         ];
 
+        let global_residency_set = desc.create_residency_set.then(|| unsafe {
```
Jasper-Bekkers, line 442: Please just use if statements

src/metal/mod.rs
```diff
@@ -166,6 +169,9 @@ pub struct AllocatorCreateDesc {
     pub device: Retained<ProtocolObject<dyn MTLDevice>>,
     pub debug_settings: AllocatorDebugSettings,
     pub allocation_sizes: AllocationSizes,
+    /// Whether to create a [`MTLResidencySet`] containing all live heaps, that can be retrieved via
+    /// [`Allocator::residency_set()`].  Only supported on MacOS 15.0+ / iOS 18.0+.
+    pub create_residency_set: bool,
```
Jasper-Bekkers, line 174: Maybe for now its a bit early, but if we do more of these platform specific settings I don't think they belong in AllocatorCreateDesc, maybe we need some platform specific traits.

Jasper-Bekkers (changes requested):

src/d3d12/mod.rs
```diff
@@ -475,28 +475,22 @@ impl MemoryType {
...
+        let is_dedicated_or_not_last_general_block =
+            !mem_block.sub_allocator.supports_general_allocations()
+                || self.active_general_blocks > 1;
+        if mem_block.sub_allocator.is_empty() && is_dedicated_or_not_last_general_block {
+            let block = self.memory_blocks[block_idx]
+                .take()
+                .ok_or_else(|| AllocationError::Internal("Memory block must be Some.".into()))?;
+
+            if block.sub_allocator.supports_general_allocations() {
+                self.active_general_blocks -= 1;
```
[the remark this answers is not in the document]
MarijnS95: Correct, that should still be caught by the `.take().ok_or_else()?` now, right?

I don't really mind which way we write this, as long as it's the same writing across all 3 duplicated backends, which it wasn't before making this code incredibly hard to refactor correctly.

Issue #274, opened 2025-06-19
Add support for external memory


inner-daemons:
In order to allocate memory that can be exported in vulkan, a `vkExportMemoryAllocateInfoKHR` struct must be passed to the `pNext` chain of the create info. unfortunately gpu-allocator doesn't allow for this. I see 3 ways to approach this:
* Add support for external memory specifically
* Add support for a custom `pNext` object for the vulkan allocator
* Add some way to access `gpu-allocator`'s internal allocator, so that the user can manage the allocation themself but still let `gpu-allocator` select the memory type and whatnot

I think option number 3 would be the most useful overall but would take the most work. I'd be interested in seeing either that or option 2 implemented here.

I'm just gonna leave this issue here until I get feedback on which approach should be taken.

Jasper-Bekkers:
I'm leaning most towards 1, that way we can also nicely unify it across the backends.

SupaMaggie70Incorporated:
> I'm leaning most towards 1, that way we can also nicely unify it across the backends.

I would say one advantage even on a single backend of 1 is that it could make use of existing algorithms to combine multiple logical allocations under the same driver allocation. That being said, I'd be in favor of opening up gpu-allocator a little even if we go for 1, but that's out of the scope of this issue.

Only thing I'd say is that some platform specific code must be written, as using fd's is slightly different than win32 handles. And there may be cases where the driver decides that some allocations must be dedicated, so this might complicate matters slightly.

If you give the green light on 1 however, I might start working on it tomorrow or some other time soon.

Jasper-Bekkers:
> Only thing I'd say is that some platform specific code must be written, as using fd's is slightly different than win32 handles. And there may be cases where the driver decides that some allocations must be dedicated, so this might complicate matters slightly.

Nicest solution I think would be to create a small side-crate like `raw-window-handle` (`raw-render-handle`) with minimal dependencies (e.g. all types are `NonNull<c_void>` etc) that we make sure is well maintained, explicitly for sharing these kind of resources around. Would be nice if it's initially part of this repository so we know that maintainership doesn't get abandoned after a while.

If that's too much effort, I would suggest to just carve out a `u64` as that can fit all pointers, handles, and file descriptors in practice.

> If you give the green light on 1 however, I might start working on it tomorrow or some other time soon.

You have my blessing.

SupaMaggie70Incorporated:
[...]

PR #275, opened 2025-06-20
Add external memory support for vulkan on windows & linux

inner-daemons:
This has been tested by my own project, though no in tree testing is there yet(not sure how I'd add that). Also, the testing is only for vulkan Linux but I suspect it would also work on Windows in its current state.

Based on discussion in #274, also going to ping @Jasper-Bekkers :)

Implementation details: the memory is only allocated with support for exporting it later. The user must manually export the memory later, though this isn't very difficult. Used as a reference was [this article from vulkan docs](https://docs.vulkan.org/guide/latest/extensions/external.html).

### DIFF
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -392,12 +392,12 @@ impl MemoryBlock {
             };
 
             // On other platforms you can't create an external capable allocator, so this would be unreachable
-            #[cfg(any(windows, all(unix, not(target_vendor = "apple"))))]
-            let alloc_info = if desc.exportable {
-                alloc_info.push_next(&mut export_info)
-            } else {
-                alloc_info
-            };
+            let alloc_info =
+                if cfg!(any(windows, all(unix, not(target_vendor = "apple")))) && desc.exportable {
+                    alloc_info.push_next(&mut export_info)
+                } else {
+                    alloc_info
+                };
 
             // Flag the memory as dedicated if required.
             let mut dedicated_memory_info = vk::MemoryDedicatedAllocateInfo::default();
--- a/src/vulkan/mod.rs
+++ b/src/vulkan/mod.rs
@@ -339,6 +339,17 @@ unsafe impl presser::Slab for Allocation {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct MemoryBlockCreateDesc {
+    pub size: u64,
+    pub mem_type_index: usize,
+    pub mapped: bool,
+    pub buffer_device_address: bool,
+    pub allocation_scheme: AllocationScheme,
+    pub requires_personal_block: bool,
+    pub exportable: bool,
+}
+
 #[derive(Debug)]
 pub(crate) struct MemoryBlock {
     pub(crate) device_memory: vk::DeviceMemory,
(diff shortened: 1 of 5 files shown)


SupaMaggie70Incorporated:
I've verified in a personal project that this code works. I'd like to get a review of it so we can hopefully get this merged! Going to ping @Jasper-Bekkers, hopefully this isn't too much of a hassle. No rush here just want to make sure you see it!

src/vulkan/mod.rs
```diff
@@ -343,9 +347,11 @@ pub(crate) struct MemoryBlock {
     pub(crate) sub_allocator: Box<dyn SubAllocator>,
     #[cfg(feature = "visualizer")]
     pub(crate) dedicated_allocation: bool,
+    pub(crate) exportable: bool,
 }
 
 impl MemoryBlock {
+    #[allow(clippy::too_many_arguments, clippy::fn_params_excessive_bools)]
```
Jasper-Bekkers, line 354: The refactor here should be to make a `MemoryBlockCreateDesc` that has these parameters.

src/vulkan/mod.rs
```diff
@@ -368,6 +375,20 @@ impl MemoryBlock {
...
             };
+            #[cfg(windows)]
+            let mut export_info = vk::ExportMemoryAllocateInfoKHR::default()
+                .handle_types(vk::ExternalMemoryHandleTypeFlags::OPAQUE_WIN32_KHR);
+            #[cfg(all(unix, not(target_vendor = "apple")))]
+            let mut export_info = vk::ExportMemoryAllocateInfoKHR::default()
+                .handle_types(vk::ExternalMemoryHandleTypeFlags::OPAQUE_FD_KHR);
+
+            // On other platforms you can't create an external capable allocator, so this would be unreachable
+            #[cfg(any(windows, all(unix, not(target_vendor = "apple"))))]
```
Jasper-Bekkers, line 386: I think on other platforms then we should fail with an error rather then silently failing to create an exportable texture.

src/vulkan/mod.rs
```diff
@@ -368,6 +375,20 @@ impl MemoryBlock {
             } else {
                 alloc_info
             };
+            #[cfg(windows)]
+            let mut export_info = vk::ExportMemoryAllocateInfoKHR::default()
+                .handle_types(vk::ExternalMemoryHandleTypeFlags::OPAQUE_WIN32_KHR);
+            #[cfg(all(unix, not(target_vendor = "apple")))]
+            let mut export_info = vk::ExportMemoryAllocateInfoKHR::default()
+                .handle_types(vk::ExternalMemoryHandleTypeFlags::OPAQUE_FD_KHR);
```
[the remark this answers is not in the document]
SupaMaggie70Incorporated: Trouble here is that they are different types, and they must exist until the allocation is actually done. So I could switch to `if cfg!()`, but I think the code would look worse (probably you would end up having an option for each type). Let me know if you still want me to make the change

SupaMaggie70Incorporated:
I think I've addressed most comments

inner-daemons:
Going to close this as I'm not getting any response.

PR #279, opened 2025-07-15
Fix all new clippy lints since Rust 1.88

MarijnS95:


.cargo/config.toml
```diff
@@ -39,7 +39,6 @@ rustflags = [
     "-Wclippy::map_err_ignore",
     "-Wclippy::map_flatten",
     "-Wclippy::map_unwrap_or",
-    "-Wclippy::match_on_vec_items",
```
MarijnS95, line 42:
```
warning: lint `clippy::match_on_vec_items` has been removed: `clippy::indexing_slicing` covers indexing and slicing on `Vec<_>`
  |
  = note: requested on the command line with `-W clippy::match_on_vec_items`
  = note: `#[warn(renamed_and_removed_lints)]` on by default

warning: `gpu-allocator` (lib) generated 1 warning
```

Fwiw.

PR #290, opened 2025-10-13
vulkan: return early if no suitable memory type found

Quandrax:
The previous allocator implementation queried for a memory type twice using identical property flags.


src/vulkan/mod.rs
```diff
@@ -819,11 +819,12 @@ impl Allocator {
 
         if memory_type_index_opt.is_none() {
             let mem_loc_required_bits = match desc.location {
-                MemoryLocation::GpuOnly => vk::MemoryPropertyFlags::DEVICE_LOCAL,
```
Jasper-Bekkers, line 822: This removal is highly dubious.

manon-traverse: I can see the logic of this change, but I don't see how this solution has any real world impact.

In the case that there is `DEVICE_LOCAL` memory, which is practically guaranteed, there is only one query. And if the extremely unlikely case that there is no `DEVICE_LOCAL` memory type we query twice, but keep the code structure simple by always trying to find a memory type with our preferred properties, and then a memory type with our required properties.

Removing that second query breaks the basic structure to save 1 query for a case that should never happen.

@Quandrax do you have any use cases where doing this extra query presents a problem?

Quandrax: I dont, no. I was just aiming for a small optimization here.

